PR #62, opened 2018-10-03
Update webpack.config.js

tharindumalshan1:


neutralino.js/webpack.config.js
```diff
@@ -4,10 +4,12 @@ module.exports = {
   entry: {
             index : './src/index.js'
         },
+  console.log(path);
```
shalithasuranga: What is the usecase of this one? @tharindumalshan1

Ditti4: To bump their number of opened pull requests. I am, too, browsing for projects to which I can contribute for Hacktoberfest and this is the second spam PR I see which was created by this user.

Issue #63, opened 2018-10-03
Change emoji in README


shalithasuranga:
Give us a star **use star emoji instead**

angelbarrera92:
I can do it!

PR #65, opened 2018-10-03
changes star instead of +1

angelbarrera92:
This pull requests solves the #63 
Thanks!


shalithasuranga:
Can you please pick another issue. This is fixed by another We checked PRs in order
 Thanks

PR #68, opened 2018-10-03
updated the thumb emoji with star emoji

pr-jli:


shalithasuranga:
it is already done @pr-jli  Please pick another issue

Issue #207, opened 2019-11-13
need app icon


acedigibits:
i have basic html5 app that doesnt need any OS APIs . Hence this repo is perfect for build as this gives low ram and disk use. 

Only requirement remaining is to have a app icon. without this it looks very unprofessional.

compelx:
I realize this has been open for a bit but I ran into this issue today and thought I'd share.
If you're on Windows you can use Resource Hacker to add a icon (.ico) to the executable. It will spit out a modified binary with the icon.

http://www.angusj.com/resourcehacker/

shalithasuranga:
Iconfile supported was added to all platforms. will be released with v1.5.0 soon

WilliamAlecBrannon:
I have updated to the current available version using neu update, and am trying to get the app icon to work. I followed the suggestion from the website to use a transparent png and it works for the taskbar, and the browser favicon, but the .exe for windows (and the other distributions, but windows is specifically the one im targetting) still shows as an exe with no set icon so it grabs windows' default application icon. According to resource hacker, the distribution has no icon associated with the exe at all. Binary version is 2.5.0 and client version is 1.2.0, and idk which the milestones are tracking

pegvin:
Currently Neutralino doesn't embed icons when building but we'll add it soon.

DevilishDante:
Hello how about this function?

AxDSan:
Still waiting for this...

tiagorangel2011:
Any update?

russellweed:
Those still waiting on official support may appreciate this workaround I've incorporated into my build process.

What you'll need to do is download an old app called [Resource Hacker](http://www.angusj.com/resourcehacker/), which is from the old Visual C++ WinForms days (because that's *still* how Windows does it). Essentially what you need to do is make your build and use ResHacker to replace the ICO file after the fact. You will also probably want to replace the metadata (Application Name, Company Name, Copyright, etc.) - this is the stuff that appears when you view the files Properties (and is used elsewhere in the OS; in my case it's used by the browser when it tries to open protocol handler links).

Of course doing this by hand got to be a pain in the ass, so I amended my `npm build` command to run a batch file to do all this for me.

```json
"build": "npm --prefix react_folder run build && neu update && neu build --release && build-reshack.bat && build-cleanup.bat"
```

[...]

Issue #185, opened 2019-10-05
Always-on-top mode


shalithasuranga:
I think we can add this as a boolean attribute to settings.json

shalithasuranga:
Added for all platforms

Issue #165, opened 2019-07-27
set window size?


Vrx8:
set fullscreen, window size, window position, window opacity at runtime, and window type, eg. borderless, there is also no official tutorial on creating executeable out of neutralino js.

shalithasuranga:
Hi. Thanks for the ideas. fullscreen support is already added to Linux version but as you mentioned it is better to have runtime manipulation of window. We will add `Neutralino.app` for those

genesy:
Also please add always on top mode for window manipulation.

shalithasuranga:
Fullscreen and always on top mode were there already :). Added `Neutralino.window.setSize(options)` for changing size properties. Allowed options are `width`, `height`, `minWidth`, `minHeight`, `maxWidth`, `maxHeight`, and `resizable`. Added opacity and positioning features to the roadmap. I will release these new things with the next release.

muditsaxena01:
Are we aware of any bugs related to setSize?
In borderless mode when I change the size (setSize) of the window, border appears after the change has been applied.

PR #223, opened 2020-02-21
Add basic app icon and always-on-top options for Linux version

momozor:
## Description
Add basic app icon option for Linux. See #207 
Add window always on top option for Linux. See #185 and #165

neu

### iconfile
- added a neutrolino.png file as default into core-linux/bin/
- ability to change picture file from core-linux/bin/app/settings.json (iconfile property)

### alwaysOnTop
- added alwaysOnTop option (using alwaysOnTop property) in core-linux/bin/app/settings.json
- use true to make the window always on top over other windows or false for the opposite of it

### iconfile
1. Build the core-linux executable
2. cd to the core-linux/bin/
3. run neutrolino and see the app icon in the DE panel

### alwaysOnTop
1. same as iconfile
2. if alwaysOnTop property set to true, try open another maximized window of other applications
like a web browser. the neutrolino window won't go into the background or 'hidden'
3. if alwaysOnTop property set to false, it will behave normally (opposite of number 2 above)

## Next steps

None.

### iconfile
- Currently only png file is known to be supported.
- At this moment, you cannot change the location of the picture. The picture ***must*** be placed inside the same directory as neutrolino executable
- by default, this property is default to "neutrolino.png" both settings.json and main.cpp

### alwaysOnTop
- by default, this property is default to false in settings.json and in main.cpp (normal behavior)


core-linux/src/main.cpp
```diff
@@ -98,14 +101,23 @@ int main(int argc, char **argv)
...
         int height = 600;
         int fullscreen = 0;
+        bool is_always_on_top = false;
+        std::string iconfile = "neutrolino.png";
         if(!options["window"].is_null()) {
             json windowProp = options["window"];
             width =  stoi(windowProp["width"].get<std::string>());
             height =  stoi(windowProp["height"].get<std::string>());
             if(!windowProp["fullscreen"].is_null())
                 fullscreen =  windowProp["fullscreen"].get<bool>() ? 1 : 0;
```
momozor: @shalithasuranga Hi. May I ask something? Why use the number as boolean representation instead of actual boolean value itself?

core-linux/bin/app/settings.json
```diff
@@ -5,6 +5,8 @@
     "window" : {
         "width" : "1000",
         "height" : "700",
-        "fullscreen" : false
+        "fullscreen" : false,
+        "alwaysOnTop" : false,
```
shalithasuranga: Could you simply use `alwaystop` or `alwaysontop` matching lowercase style of other settings

core-linux/src/main.cpp
shalithasuranga: Ah sure.. that's because webview's method accepts int guess. So we just simply used int :)

core-linux/bin/app/settings.json
momozor: sure

core-linux/src/main.cpp
momozor: oh i see. thanks for the explanation

momozor:
> Awesome work +1 I really like if we could add this to other platforms as well when we do next release.. Could you please check inline comments

I don't have access to either Windows or Mac. Sorry...

inline comments?

shalithasuranga:
Thanks.. it's merged

momozor:
:D

PR #243, opened 2020-03-12
add borderless window option for linux

### LINKED ISSUE #242: Can generate widgets apps?
**Expected Behavior**  

I would like to be able to create a widget application that has no window borders. This way you could create a Dock or a panel
  

**Actual Behavior**  

At the moment this functionality does not exist or at least it is not detailed. Is this possible?

**Specifications**  
* NeutralinoJs Version: 1.0.8
* Platform: Linux

momozor:
## Description
add borderless window option for Linux 

## How to test
- set borderlesswindow property to true in settings.json to enable borderless/un-decorated top window
- false for the opposite

Fixes #242

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ To create Neutralino developer environment in your pc, install these tool kits.
 
 Install GTk and webkit libraries
 
+#### On Debian/Ubuntu based OS
 ```bash
 $ sudo apt-get install libgtk-3-dev
 $ sudo add-apt-repository ppa:webkit-team/ppa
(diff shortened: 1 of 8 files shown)


shalithasuranga:
@momozor  I think we can try to add newly added features to other platforms as well than adding more and more to linux version. We can do a release I guess. I will also try to work on those newly created issues

momozor:
sure

quybeans:
is borderless coming to MacOS and Window as well, can't wait 🔥🔥

Issue #159, opened 2019-06-06
Windows App Title with Spaces


Prid13:
The `<title>` tag is not working on Windows app title bar, as it only shows the value of `appname` in `settings.json`. No problem with that. However, as soon as I type in an appname with **spaces**, the Windows Neutralino.js app stops working and prompts a Save As dialog for a JSON file with the appname, and an error saying "Navigation to the website was canceled.".

![image](https://user-images.githubusercontent.com/27897747/59059680-dcf91600-889f-11e9-924f-382ae947ff46.png)

![image](https://user-images.githubusercontent.com/27897747/59059707-eaae9b80-889f-11e9-8400-4521b1103304.png)

momozor:
On linux version, it behaves a bit weird 
![lop](https://user-images.githubusercontent.com/24475030/75299328-b9339100-586f-11ea-967c-e52f694bb6e5.png)

ainar:
Same behavior as in Linux when opening in browser on Windows.

shalithasuranga:
It was fixed in Windows via #257 . We need to add into other platforms as well

shalithasuranga:
Now `window.title` can be used with spaces etc for the title. and appname should act as app id without space etc

PR #257, opened 2020-04-07
fix url encoding in routes

ainar:
This is a fix for bug #159.

No special character including spaces is decoded for now in routing. This lays to the default route returning `{"message":"Neutralino"}`.

My solution is to [decode paths](https://en.wikipedia.org/wiki/Percent-encoding) before routing.

### DIFF
--- a/core-windows/src/functions.cpp
+++ b/core-windows/src/functions.cpp
@@ -55,5 +55,23 @@ namespace functions {
         return s;
     }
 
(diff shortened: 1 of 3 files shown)


kasvith:
LGTM

shalithasuranga:
Thanks .. we need to add into other platforms as well before releasing :+1:

PR #393, opened 2021-03-18
allow access to clipboard

akvadrako:
## Description

Since Neutralino apps are local apps, they should be able to paste.

## How to test it

```html
    <div>
        <h2>Clipboard Test</h2>
        <input id='output' />
        <input id='paste' type='button' value='paste' />
        <input id='copy' type='button' value='copy' />
        <script>
            var $ = x => document.querySelector(x)
            var out = $("#output")

            $('#paste').onclick = () => {
                out.focus()
                out.select()
                document.execCommand("paste")
            }
            
            $('#copy').onclick = () => {
                out.select()
                document.execCommand("copy")
            }
        </script>
    </div>
```


shalithasuranga (approved):
Thanks :+1:

Issue #141, opened 2019-01-30
FreeBSD support


Peter2121:
Please, add FreeBSD support.

momozor:
Assuming you already installed `gnome3`, `xorg`, and `emulators/linux_base-c7`..and enabled the linux compatibility modules, you should be able to run the neutrolino-linux build. I did a test on this, but it returned this error `error while loading shared libraries: libwebkit2gtk-4.0.so.3: cannot open shared object file: No such file or directory`.

Peter2121:
Hmm... Why do one need to install gnome3? Can you build it on Linux using recent clang? If not - probably installing gcc from package could improve your build on FreeBSD...

momozor:
No you don't need gnome3 for that, indeed.

momozor:
> I also tried to build the linux version on freebsd, but it failed and complained about 'incomplete types'

changing C++ standard to -std=c++17 should fix this error

shalithasuranga:
Closing this because now FreeBSD support is added. Thanks @lhondareyte for the pull request.

PR #482, opened 2021-06-19
Fix compilation on FreeBSD

lhondareyte:
## Description
Fix building on FreeBSD

## Changes proposed
Adding script `build_freebsd.sh` and set various `#define (__FreeBSD__)`


shalithasuranga (approved):
Thank you very much.. Now we can add the FreeBSD support for CLI as well when we release this version. I  will check with a vm.

lhondareyte:
Thanks for your merge. Perhaps, you could close issue #141

PR #489, opened 2021-06-26
Fix more missing features for FreeBSD

lhondareyte:
To build neutralino on FreeBSD, install the following packages
```
pkg install -y gtk+-3.0 webkit2gtk-4.0 glib-2.0 appindicator3-0.1
sh build_freebsd.sh
```

### DIFF
--- a/api/app/app.cpp
+++ b/api/app/app.cpp
@@ -1,6 +1,6 @@
 #include <iostream>
 
-#if defined(__linux__) || defined(__APPLE__)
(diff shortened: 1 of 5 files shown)


shalithasuranga:
Hey.. @lhondareyte  Thank you very much. I added a build job in GitHub actions for FreeBSD. Just noticed the commands you mentioned about are failing in the VM instance: https://github.com/neutralinojs/neutralinojs/runs/2923997736#step:3:1381

Could you please check when you have some time :)

PR #499, opened 2021-07-02
Fedora/RHEL based Linux error fix and install guide

AOSXAP:
## Description

Updated readme for "Application was terminated." error (on Linux) and guide to install on Fedora/RHEL based distros,


shalithasuranga (approved):
Thanks :+1:

PR #529, opened 2021-07-24
Removed unused interval parameter

input-variable:
## Description
Removed an unused interval parameter. No references anywhere in the codebase, seems like it was added in error.

## Changes proposed
 - Removed the interval parameter in the function declaration.
 - Removed the "10000" parameter in the function call.

## How to test it
- Check if the pinging system still works?

## Next steps

None.

## Deploy notes

None.

### DIFF
--- a/server/ping.cpp
+++ b/server/ping.cpp
@@ -9,7 +9,7 @@ using namespace std;
 
 bool isActive = true;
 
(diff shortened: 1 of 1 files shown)


shalithasuranga (approved):
Thank you :+1:

Issue #531, opened 2021-07-26
[SUGGESTION] Allow file filtering in Open Dialog


hypherionmc:
**Expected Behavior**  
Currently. the showDialogOpen dialog will show all files in all folders. It would be really nice to have the ability to add a filter, so that it only shows certain extensions.
  
The api function could be:
```javascript
let response = await Neutralino.os.showDialogOpen({
  title: 'Select a file',
  isDirectoryMode: false,
  filter: ["txt", "bmp", "myextension"]
});
```

As far as I know, all operating systems include a filter feature for these dialogs, so it shouldn't be impossible to add

hypherionmc:
I'm busy adding this, so I'll do a PR for it :)

shalithasuranga:
@hypherionmc Thanks this is a great feature for open dialog :+1:

hypherionmc:
I thought this was a feature already, until my app update yesterday and noticed it wasn't. So far I got the filter working on Linux and Windows, but mac is still a bit tricky

shalithasuranga:
Maybe Apple script's `of type {"txt", "bmp"}` syntax will work?

hypherionmc:
It does, but only for common extensions apparently. For example: `of type {"png"}` works, `of type {"toml"}` does not

hypherionmc:
For more context, this is on macOS Big Sur 11.4. Not sure if this is an OS issue, or if just a general bug

hypherionmc:
Okay so doing `type of {"", "toml"}` works fine. Will probably have to do it that way then

hypherionmc:
Found another Linux related bug. When calling the zenity command, it throws a warning from GTKDialog. This warning, is then passed along with the file path to selectedEntry. I added a fix for this as well

hypherionmc:
Also found a similar bug in MacOS. The macOS one just adds a new line to the result returned when you open the file. I should be able to PR this tomorrow. Just installing Visual Studio so I can test the windows build, unless there is another build system I can use?

shalithasuranga:
Thanks, @hypherionmc for the pull request. It's merged and I will release this feature with the next version.
I did a few adjustments to your changes.
- moved the ending newline replacement logic to `excecCommand`
- Displayed `.` before the file extension name on Windows. So on Windows, we can see selectors like `.cpp files` etc.
- Did some minor changes to variable names etc.

hypherionmc:
[...]

PR #534, opened 2021-07-27
Added support for file extension filter in File Select dialogs and Fixed some Linux/MacOS bugs

hypherionmc:
## Description

This adds a new api function to `os.showDialogOpen` to allow users to add file filters to the dialog so that only certain file types can be selected as requested in #531 and also fixes some formatting issues on Linux and Mac regarding the values returned from os::ExecuteCommand()

## Changes proposed

 - Added a new setting to `os.showDialogOpen` called filter. (Small example below)

## How to test it

 - Call `os.showDialogOpen` with `isDirectoryMode` set to `false` and add `filter: ["png", "jpg", "*"]` to the settings. So the call should look like this:

```javascript
let response = await Neutralino.os.showDialogOpen({
  title: 'Select a file',
  isDirectoryMode: false,
  filter: ["png", "jpg", "*"]
});
console.log(`You've selected: ${response.selectedEntry}`);
```

## Next steps

None.

## Deploy notes

None.

### DIFF
--- a/api/os/os.cpp
+++ b/api/os/os.cpp
@@ -211,7 +211,19 @@ namespace controllers {
             command += " --title \"" + input["title"].get<std::string>() + "\"";
         if(!input["isDirectoryMode"].is_null() && input["isDirectoryMode"].get<bool>())
             command += " --directory";
-        output["selectedEntry"] = os::execCommand(command);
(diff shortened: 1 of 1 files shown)


hypherionmc:
Just a note, I tested this function on Linux, Mac and Windows. They worked fine in testing

shalithasuranga (approved):
Thank you very much.. I will do some tweaks for the change and let you know

PR #679, opened 2021-10-30
Improve os.execCommand API

shalithasuranga:


appsdevpk:
is it possible to add context to execCommand, like i have a folder named projects inside my app and i want to run os commands inside that folder.

shalithasuranga:
Hey.. currently there is no option to pass the working directory. Maybe we can add it as a feature in the future. However, there are some workarounds for that,

- Using `cd` before running commands (Maybe adding into a method like `addCwdCmd()`)
- Using `NL_CWD + your folder` as an argument to your commands (if supported)

Issue #486, opened 2021-06-26
Support for older versions of Windows (7, 8, and 8.1)


shalithasuranga:
- Test on different windows versions
- Release a 32-bit version

pathange-s:
I am interested in testing the functionalities of neutralinojs on windows 7. I am setting up a VM locally and then install windows 7. I will try to cover major functionalities. Hope to get inputs when ever required.

pathange-s:
Windows 7: Node version 13. 14.0 LTS is the last installer that works on Window 7. But, in the [requirements](https://www.npmjs.com/package/@neutralinojs/neu), its stated that node version v14.x.x or greater is required. Also tried to verify it by testing it on a windows 7 vm. So this issue boils down to making neutralinojs compatible with older node versions also.

![image](https://user-images.githubusercontent.com/76244077/160852328-a9303215-4ea0-4a7d-9f37-7ca9590c6e2a.png)

![image](https://user-images.githubusercontent.com/76244077/160984353-26de3103-dd10-4f2c-addd-d95169bf57e0.png)

Tried running `node <file_name>` to check if this error persists in every file, but few files ran successfully.

- modules/bundler.js
- modules/filewatcher.js
- commands/build.js
- commands/run.js

These files threw the same `SyntaxError: Unexpected token ','`

pathange-s:
Its been long, took me a little while to setup required environment and test `nue`'s fucntionality on Windows OS. I was able to draw a report for Windows 7, will work on other OS eventually, once sorted on one Windows's version. [Here's](https://docs.google.com/document/d/1sXq8hfD1lQSgDVhCkaNxtqTygGhzedTqEqH9MahYV-U/edit?usp=sharing) the doc file for the same. I was struck at the last described error. Will be glad if any Windows's expert could help me out in adding the executable to the DEP list.

I hope adding the exe file to DEP list should work, not sure though. I guess it was being blocked by internal Windows system. Any other ways to solve `App Crash` error is encouraged to drop below.

cc: @DEVLOPRR I see you tried running neu app on windows. Hope you could help me.

pathange-s:
![image](https://user-images.githubusercontent.com/76244077/161949815-2d939696-3c98-43f9-a7a1-84a20e94cb39.png)
(The pixelating issue is because I am running a VM with some resolution issues)

Ran app successfully with some modifications in source code. Y'all can expect support for, atleast Windows 7 soon :)

pathange-s:
[...]

PR #804, opened 2022-02-17
#486: Call SetProcessDpiAwarenessContext only when available

tshm:
## fix launch failure issue on win7/win8.1

The failure is caused by the fact that SetProcessDpiAwarenessContext function
does not exist in older windows before windows 10.
This fix is merely avoiding calling the function if it does not exist in user32.dll.
Ultimate fix should include calling equivalent functions in older windows.

## Changes proposed

 - check the existance of the function and call it if available.  Ignore otherwise.
 - split the logic into separate method.  So that it can be extended to support
   older Windows in future.

### DIFF
--- a/lib/webview/webview.h
+++ b/lib/webview/webview.h
@@ -1238,7 +1238,7 @@ public:
       m_window = *(static_cast<HWND *>(window));
     }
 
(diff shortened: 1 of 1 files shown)


shalithasuranga (approved):

shalithasuranga:
Thank you very much for the contribution :tada:

PR #851, opened 2022-04-02
Prevent cluttering of root folder when building on Windows

rkharsan:
## Description
Currently, building on Windows litters the project root with temporary object files. While this is not a problem as object files are ignored, it is slightly annoying to work with when using VS Code.

## Changes proposed
 - Add the `/Fobin/` flag to Windows build options. This puts the object files in the `bin/` folder.

## How to test it
 - Run build using `python scripts/bz.py`. All temporary object files should be in `bin/`.

## Next steps
None.

## Deploy notes
None.

### DIFF
--- a/buildzri.config.json
+++ b/buildzri.config.json
@@ -56,6 +56,7 @@
         "windows": [
             "/EHsc",
             "/Os",
+            "/Fobin/",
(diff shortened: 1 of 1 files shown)


shalithasuranga (approved):
Thanks for the contribution :tada:

PR #852, opened 2022-04-03
Update some libraries to newest versions

rkharsan:
## Description

Updated some libraries. While no major features/bugfixes were introduced, keeping dependencies up to date is a good idea for the future.

Also reorganized `lib/asio` while updating (see below).

## Changes proposed

Updated the following libraries to latest versions:
 - `lib/clip`
 - `lib/filedialogs`
 - `lib/platformfolders` - includes an additional function `getStateDir()`. This might come handy in the future.
 - `lib/tinyprocess`
 - `lib/asio` - removed unnecessary folders and files. Also removed `include` folder and moved files directly to `lib/asio`. Updated `buildzri.config.json` to new include path.

## How to test it

 - Build using `python scripts/bz.py`. No build errors.
 - Run spec/tests. (Tested on Windows 10)

## Next steps

None.

## Deploy notes

None.


rkharsan:
> There are changes in numerous files. May I know why ? Are these generated / replaced while updating packages ?

The libraries in `lib/` are copied directly from the releases available at their respective websites/Github releases. I extracted the latest version of all libraries. Also the huge line count is because I removed unnecessary files from `lib/asio`. If you go to `lib/asio` today, you'll notice tons of files in `lib/asio/src`, but these are not necessary for us (they contain tests and tools for developing ASIO). So I removed these.

After removing excess stuff, I was left with just a single directory, `include` inside `lib/asio`, which contained all header files. So I removed the extra folder and updated the include path from `lib/asio/include` to `lib/asio` directly.

Hope this explains why this PR has such a huge line count, despite not much changing.

PR #853, opened 2022-04-04
Update README.md

NightyLighty:


DEVLOPRR:
Thanks For Fixing The URL, but always make sure to add a detailed description about what has changed. like you can add "Fixed ReleaseZri URL"

shalithasuranga (approved):
Thanks :+1:

PR #872, opened 2022-04-23
Fix broken test suite (Github Actions)

rkharsan:
## Description
Currently, the checkout action only fetches a single commit. When the tests run on a branch other than main, then build fails the as `git log ...` command in `bz.py` cannot find the main branch locally (as it does not exist on the CI machine).

This can be seen in:
- https://github.com/neutralinojs/neutralinojs/runs/6112616322?check_suite_focus=true
- https://github.com/neutralinojs/neutralinojs/runs/6141133611?check_suite_focus=true

If you check the output of the build step in theses links (scroll up), you can see that the error message by `git log` is mixed with the build command, causing the build to fail.

## Changes proposed
Fetch all branches by using `fetch-depth: 0`.

EDIT: Also get commit hash from `origin/main` instead of `main`.

## How to test it
Try running any CI tests after this change is merged.

## Next steps
None.

## Deploy notes
None.

### DIFF
--- a/.github/workflows/test_suite.yml
+++ b/.github/workflows/test_suite.yml
@@ -8,8 +8,6 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v2
-      with:
-        fetch-depth: 0
 
     - name: Install Libraries
       run: |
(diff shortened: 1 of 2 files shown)


shalithasuranga (approved):
Thanks for the contribution :tada: Please note that I've done a few tweaks before merging ;)

PR #887, opened 2022-05-09
Update filesystem.getStats(path) to fetch last modified and created time

### LINKED ISSUE #876: Add time related information to `filesystem.getStats` result.
Hello,

with this issue I would like to propose an extension to the return object of the `filesystem.getStats` method. It currently only returns the information about the checked item being a file or directory and its size. In certain scenarios it would be great to also have access to the file creation and last file modification dates.

So please consider extending the return object from the current state of

```javascript
{
  isDirectory: false,
  isFile: true,
  size: 11
}
```

to something like

```javascript
{
  isDirectory: false,
  isFile: true,
  size: 11,
  createdAt: 1650956224,
  modifiedAt: 1650956224
}
```

where both times are numbers represent the milliseconds in unix time.

Thank you very much for considering this feature request. Keep on the great work on Neutralino!

Sincerely,
Happy Striker

pathange-s:
Fixes #876 
[WIP]

Update `filesystem.getStats()` function to return last modified time and created time in seconds when called.

To test call the following function:

```
async function displayGetStats() {
    try{
    let stats = await Neutralino.filesystem.getStats('path_to_file');
    console.log('Stats:', stats);
    }
    catch(e){
        console.log(e);
    }
}
displayGetStats();
```

### DIFF
--- a/api/filesystem/filesystem.cpp
+++ b/api/filesystem/filesystem.cpp
@@ -14,6 +14,12 @@
 #include <atlstr.h>
 #include <shlwapi.h>
 #include <winbase.h>
+#define WINDOWS_TICK 10000000
+#define SEC_TO_UNIX_EPOCH 11644473600LL
+long long WindowsTickToUnixMilliSeconds(long long windowsTicks)
+{   long long unixTime = (windowsTicks / WINDOWS_TICK - SEC_TO_UNIX_EPOCH)*1000;
+    return unixTime;
+}
 #endif
 
 #include "lib/json/json.hpp"
@@ -118,6 +124,8 @@ fs::FileStats getStats(const string &path) {
         fileStats.size = statBuffer.st_size;
         fileStats.isFile = S_ISREG(statBuffer.st_mode);
         fileStats.isDirectory = S_ISDIR(statBuffer.st_mode);
+        fileStats.createdAt = statBuffer.st_ctime*1000;
+        fileStats.modifiedAt = statBuffer.st_mtime*1000;
     }
 
     #elif defined(_WIN32)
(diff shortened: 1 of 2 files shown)


pathange-s:
I created the same file changes on a pull request in my forked repo. There, the checks run fine. Check them [here](https://github.com/pathange-s/neutralinojs/pull/3)

api/filesystem/filesystem.cpp
```diff
@@ -27,6 +30,13 @@ using namespace std;
 using json = nlohmann::json;
 
 namespace fs {
+    
+#if defined(_WIN32)
+// From: https://stackoverflow.com/a/6161842/3565513
```
shalithasuranga, line 35: Hey @pathange-s  thank you very much for sending this pull request. Could you please let me know if you are okay with adding this comment to your helper function implementation? We always love to give credit to people who provide these free solutions. If you found this solution from another source (or by yourself) Please let me know too ;) I think after discussing this we can merge this pull request :tada:

pathange-s: Yes! I am okay with the following comment. Infact, loved the idea of mentioning it. Also, I tweaked it to make it suitable for our need.

shalithasuranga: Great! :tada:

shalithasuranga: Hmm.. I was thinking a bit about `createdAt` since it's not really the creation time on Linux (rather an inode change timestamp). However, let's merge this one -- we can think about it before the next release ;)

pathange-s: https://github.com/neutralinojs/neutralinojs.github.io/pull/204/files#diff-987ec3372f278ebc4449566c57aea73655f1a17e1d8840e875ae0a9ec70cad52R186
"It depends on the Operating system"

PR #920, opened 2022-06-06
Added ability to set default filename on showSaveDialog()

kodaloid:
This change modifies the portable-file-dialogs.h library to add the ability to set a default filename. It also extends the os.showSaveDialog method to allow users to specify that filename, and the default folder in their JavaScript parameters.

I've tested the modifications to portable-file-dialogs.h by creating a separate Win32 app, referencing the library, and using it there. The filename comes through correctly without issue.


kodaloid:
Closed this as there is a better way to do what's needed without pulling a new version of portable-file-dialogs.h

Issue #930, opened 2022-06-13
Display Window Flashes Although Hidden: true


danidre14:
**Describe the bug**
When you launch a Neu application (either `neu run` or double clicking the `.exe` after a `neu build`), even though in the config `"hidden": true`, the window flashes on screen for 1 second then disappears, causing unexpected behaviour.

**To Reproduce**

1. Go to `neutralino.config.json`

2. Set `"hidden": true`
![example_hidden_true](https://i.imgur.com/nS3svWa.png)

3. Go to `resources/js/main.js`

4. Add code to bottom of file under `Neutralino.init` to show window after 2 seconds:
```js
setInterval(()=>Neutralino.window.show(),2000);
```
![example_show_window](https://i.imgur.com/mAEK9Cb.png)

5. Run application (with `neu run` or executable)

6. See error

**Expected behavior**
The app should begin, then two seconds later, the window should appear.

**Resulting behavior**
The app window flashes on screen, then vanishes, then two seconds later, appears once more.

**Screenshots/Screen recordings**
If applicable, add screenshots/screen recordings to help explain your problem.
![neu_windows_screen_recording](https://i.imgur.com/a8qpMnA.gif)

**NOTE**

I believe this is a **Windows 10** specific problem. I tested the same on **Linux** but no flash was seen and it worked as expected.

**Specifications**  
 - OS: Windows 10 x64
 - Neutralinojs version: [e.g. v4.6.0]
 - Neutralinojs client library version: [e.g. v3.5.0]
 - Neutralinojs CLI version: [e.g. v9.2.0]

*Note: you can enter the `neu version` command to find these details*

**Additional context**
When I launch the app I want it to stay hidden until a certain promise is resolved (here I just did the 2 second setInterval for testing purposes). The flash on Window's versions is very negatively received.

Blatman:
Just wondering - is this a duplicate of #900 ?

Blatman:
Maybe it is an idea to reference the other issue then both can be closed when problem is fixed.

shalithasuranga:
This was fixed with #1061 and will be released with the next framework release. Thank you :tada:

PR #950, opened 2022-07-16
Fix popup issue in windows when hidden set to true.

kodaloid:
## Description
On Windows if you launch Neutralino with `hidden: true` in the config. A window appears for a split second before disappearing. Quite a few people have expressed a wish to have this changed so that the window never appears with this setting, so this pull request introduces a solution that should work given the conditions.

I tested a lot of solutions before hand. Ideally using SW_HIDE when the window is first created would be the solution, however if you do that, the WebView component always fails to load, with a corrupt connection to it's developer tools.

## Changes proposed
The changes made in this pull request are all scoped specifically for Windows using `#if defined(_WIN32)` when needed:

- Set the X coordinate to a crazy high number when CreateWindow is called.
- Change the ex window style to WS_EX_TOOLWINDOW so that no icon appears in the taskbar.
- Provide two helper methods called `window::isFakeHidden()` and  `window::undoFakeHidden()`.

If the window didn't need to hide, or show() is called, the helper methods above provide a way to show the window again to the user.

## How to test it
Build and use the new binary, everything should work as normal, however when you set hidden to true, there will no longer be a window flashing before it hides.

### DIFF
--- a/api/window/window.cpp
+++ b/api/window/window.cpp
@@ -141,6 +141,24 @@ bool isVisible() {
     #endif
 }
 
+bool isFakeHidden() {
+    //ensureOnScreen
+    RECT rect = { NULL };
+    if(GetWindowRect( windowHandle, &rect)) {
+        return rect.left > 9999;
+    }
+    return false;
+}
+
+void undoFakeHidden() {
+    #if defined(_WIN32)
+    ShowWindow(windowHandle, SW_HIDE);
+    SetWindowLong(windowHandle, GWL_EXSTYLE, nativeWindow->m_originalStyleEx);
+    SetWindowPos(windowHandle, NULL, 10, 10, 0, 0, 0x0004 | 0x0001);
+	ShowWindow(windowHandle, SW_SHOW);
+    #endif
+}
+
 void show() {
     if(window::isVisible())
         return;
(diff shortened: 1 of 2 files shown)


pegvin:
Fix for #930

pegvin:
@shalithasuranga this pr was a bug fix, please review it.

shalithasuranga:
Hey.. all, @kodaloid, thanks for sending this, I was planning to merge this pull request before this month's release but it looks like there is something to check further in the webview library. Let's add this solution as the first task after the release :tada: Thanks

kodaloid:
> Hey.. all, @kodaloid, thanks for sending this, I was planning to merge this pull request before this month's release but it looks like there is something to check further in the webview library. Let's add this solution as the first task after the release 🎉 Thanks

No problem, will take a browse through the other issues to see if there's anything else I can contribute too.

danidre14:
Any update on when this will be merged?

kodaloid:
I imagine this may need tweaking as the most recent version of Neutralino has changed a few things. Have we found an alternative solution to the issue? If not, I'll book some time in to rework the pull request to see if we can move forward with this.

Koda

kodaloid:
Closing this so that I can rework the patch to fit the latest release.

PR #1010, opened 2022-10-11
Fix build for Apple Silicon

denisw:
## Description

The Apple Silicon (`arm64`) version of the Apple macOS SDK forces all occurrences of `obc_msgSend` to be cast so that the parameter and return types match the call.

https://www.mikeash.com/pyblog/objc_msgsends-new-prototype.html

The Neutralino.js codebase does these `obc_msgSend` casts in most places already, but they are missing in `tray.h`. As a consequence, the code does not compile with `macos-arm64` as the target.

## Changes proposed

Change all `objc_msgSend` calls in `tray.h` to cast the function to the right type first.

```c
// Before:
objc_msgSend(menu, sel_registerName("autorelease"));
objc_msgSend(menu, sel_registerName("setAutoenablesItems:"), false);

// After:
((id (*)(id, SEL))objc_msgSend)(menu, sel_registerName("autorelease"));
((id (*)(id, SEL, bool))objc_msgSend)(menu, sel_registerName("setAutoenablesItems:"), false);
```

## How to test it

Right now, `scripts/bz.py` always does an Intel build, even when run on an Apple Silicon machine (see [codrezri/buildzri#7](https://github.com/codezri/buildzri/issues/7)). But you can force an Apple Silicon build by editing the script:

1. Open `scripts/bz.py` in your editor.
2. Find the `get_arch()` function.
3. Change `use_mac_rosetta = True` to `use_mac_rosetta = False`.
4. Run `python scripts/bz.py`.

## Notes

The formatting of `tray.h` seemed quite inconsistent (e.g., indentation varied between 2 and 4 spaces, often within the same function). As two-space indentation and four-space continuaition lines seemed like the most common convention in the file, I changed all functions I touched to be formatted the same.

## Next steps

- We should allow doing an Apple Silicon build without editing the build script. 
- If possible, we could pre-compile Apple Silicon binaries through GitHub Actions.

### DIFF
--- a/lib/tray/tray.h
+++ b/lib/tray/tray.h
@@ -178,31 +178,53 @@ static id statusItem;
 static id statusBarButton;
 
 static id _tray_menu(struct tray_menu *m) {
-    id menu = objc_msgSend((id)objc_getClass("NSMenu"), sel_registerName("new"));
-    objc_msgSend(menu, sel_registerName("autorelease"));
-    objc_msgSend(menu, sel_registerName("setAutoenablesItems:"), false);
+  id menu = ((id (*)(id, SEL))objc_msgSend)(
+          (id)objc_getClass("NSMenu"),
+          sel_registerName("new"));
 
-    for (; m != NULL && m->text != NULL; m++) {
-      if (strcmp(m->text, "-") == 0) {
-        objc_msgSend(menu, sel_registerName("addItem:"),
-          objc_msgSend((id)objc_getClass("NSMenuItem"), sel_registerName("separatorItem")));
-      } else {
-        id menuItem = objc_msgSend((id)objc_getClass("NSMenuItem"), sel_registerName("alloc"));
-        objc_msgSend(menuItem, sel_registerName("autorelease"));
-        objc_msgSend(menuItem, sel_registerName("initWithTitle:action:keyEquivalent:"),
-                  objc_msgSend((id)objc_getClass("NSString"), sel_registerName("stringWithUTF8String:"), m->text),
-                  sel_registerName("menuCallback:"),
-                  objc_msgSend((id)objc_getClass("NSString"), sel_registerName("stringWithUTF8String:"), ""));
+  ((id (*)(id, SEL))objc_msgSend)(menu, sel_registerName("autorelease"));
+  ((id (*)(id, SEL, bool))objc_msgSend)(menu, sel_registerName("setAutoenablesItems:"), false);
 
-        objc_msgSend(menuItem, sel_registerName("setEnabled:"), (m->disabled ? false : true));
-          objc_msgSend(menuItem, sel_registerName("setState:"), (m->checked ? 1 : 0));
-          objc_msgSend(menuItem, sel_registerName("setRepresentedObject:"),
-            objc_msgSend((id)objc_getClass("NSValue"), sel_registerName("valueWithPointer:"), m));
+  for (; m != NULL && m->text != NULL; m++) {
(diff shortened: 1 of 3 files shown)


lib/tray/tray.h
```diff
@@ -178,31 +178,53 @@ static id statusItem;
...
-          objc_msgSend(menu, sel_registerName("addItem:"), menuItem);
-
-          if (m->submenu != NULL) {
-            objc_msgSend(menu, sel_registerName("setSubmenu:forItem:"), _tray_menu(m->submenu), menuItem);
+  id menu = ((id (*)(id, SEL))objc_msgSend)(
+          (id)objc_getClass("NSMenu"),
+          sel_registerName("new"));
+
+  ((id (*)(id, SEL))objc_msgSend)(menu, sel_registerName("autorelease"));
+  ((id (*)(id, SEL, bool))objc_msgSend)(menu, sel_registerName("setAutoenablesItems:"), false);
```
JannikGM, line 186: All of these code changes are pretty ugly.

Why is this necessary in the first place, and if it's really necessary, wouldn't it make sense to put it in a wrapper (or even a macro), so it's not as ugly as it is right now?

JannikGM: Looks like `OBJC_OLD_DISPATCH_PROTOTYPES` was responsible to use the legacy variadic function.
Googling around I found https://github.com/Boscop/web-view/issues/292 and it looks like this is indeed the solution: https://github.com/Boscop/web-view/pull/294/files

Ugly. But okay then.

JannikGM: @denisw You should be removing `OBJC_OLD_DISPATCH_PROTOTYPES` from https://github.com/neutralinojs/neutralinojs/blob/f09d63be9d1011f8d64a142743bd2d5f48723d40/buildzri.config.json#L82 then

Otherwise this looks good to me.

denisw: Alight, I'll remove `OBJC_OLD_DISPATCH_PROTOTYPES` then.

And yes, I agree that it's super ugly, but I found no general way to simplify it or wrap it in a macro, and there seems no way around it - short of using Objective-C. 😓

JannikGM: I assume the proper solution would be to use another high level abstraction which casts internally. But I'm not an Objective-C / Apple developer, so I wouldn't know what exists already.

denisw: That high-level abstraction is exactly what Objective-C is, I'm afraid. There seems to be no way around the manual casts when sending messages from C, and because every method signature requires a different cast, it's hard to abstract.

denisw: `OBJC_OLD_DISPATCH_PROTOTYPES` is now removed. The build seems to still succeed.

shalithasuranga:
Hey @denisw thanks for sending this PR, and thanks @JannikGM  for reviewing :) .. @denisw will you be able to apply changes by keeping the original code formatting (as in https://github.com/neutralinojs/tray)? Right now, we can't easily identify changes I guess. Once done, we can merge this soon as it solves a crucial issue :tada:

shalithasuranga:
Hey, @denisw Thank you so much for sending this -- we've merged #1025 which was created based on these changes and other improvements added by @JannikGM

shalithasuranga:
Hey @JannikGM Thanks for your finding :) I'll add that change

shalithasuranga:
Added via [this commit](https://github.com/neutralinojs/neutralinojs/commit/68ece084decf99286c12cfafd07a368c20e93ac3) :tada:

PR #1025, opened 2022-11-10
Add cross-compile support for macOS and create universal binary (ARM for Apple Silicon M1 / M2 + x86_64 for Intel based mac)

JannikGM:
Based on #1010, but rebased to latest master.
Therefore it's reviewed best commit-by-commit.

**Note: I did not test if the artifacts are good because my own fork is missing the GitHub tokens.**

## Description

I've added cross-compilation to macOS, so we can compile for x64 and arm64 (Apple Silicon aka. M1 / M2) macs regardless of the CI platform.
This also makes CI more robust, as migration to arm based CPUs won't break the CI.

Additionally, we'll bundle the arm64 and x64 binary into a single universal binary (of larger size), so the user doesn't have to worry about picking the appropriate binary (which avoids Rosetta, if possible).

## Changes proposed

- Remove non-sensical `use_mac_rosetta` from build-script, because rosetta is an emulator at runtime, not a build flag
- Add arch argument to the macOS compiler line, so we select the target arch for cross-compilation (Documented here: https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary#Update-the-Architecture-List-of-Custom-Makefiles)
- Expose `target_arch` command line argument as a new flag. It takes the arch "short-name" (more of a portable and unique identifier really). Also migrated to Python `argparse` (Documented here: https://docs.python.org/3/library/argparse.html); the option uses the arch identification which is already used for `short_names`
- Add custom logic to bring the generic arch "short-name" into a format that's suitable for the compiler (which must be done per platform/compiler, so it's inherently tied to the python script). Because each compiler has their own preference for arch names (sometimes even depending on the context) it's not enough to have a generic substitution in the shell script.
- Trigger an error when trying to cross-compile on another OS. This should be possible, but it's untested, therefore blocked. We'll have to add more boilerplate to bring the arch names into whatever the compiler expects.
- Add GitHub Actions support to actually use the new command line argument, so we build 2 macOS binaries (x64 and arm64)
- Because GitHub Actions runs on macOS, we can build a universal binary using the `lipo` tool (Documented here: https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary#Update-the-Architecture-List-of-Custom-Makefiles)

## How to test it

- Run `./scripts/bz.py` on macOS, it should build for the host architecture
- Run `./scripts/bz.py --target_arch x64` on macOS, it should build for intel based mac
- Run `./scripts/bz.py --target_arch arm64` on macOS, it should build for apple based mac
- Run `./scripts/bz.py --target_arch x64` on a non x64 CPU and it should report an error (unless you are on macOS)
- Run `./scripts/bz.py --target_arch armhf` on a non armhf CPU and it should report an error (unless you are on macOS)
- Run `./scripts/bz.py --target_arch arm64` on a non arm64 CPU and it should report an error (unless you are on macOS)
- Push release, should have 3 macOS builds now (x64, arm64, universal)

## Next steps

Tests have not been touched, except forcing x64 build. I wasn't sure how the test script picks which neutralino binary to run (if any?), so it didn't seem useful to build both versions.
I also don't build the universal app during the test as it's not affected by code.

Some flags (for example `-host-arch` and `-arch`) in the Windows specific portion look very bad.
The flags added by the script and through the config JSON are also closely tied to the compiler in question.
Code should be refactored some more, so we can also build for Windows ARM from other archs.
However, I believe the build script needs major refactors in the future; I also wonder if it should be replaced with something like CMake which would solve many challenges while reducing code size (at expense of adding a very common dependency for C / C++ projects).

## Deploy notes

Should ensure that the 3 macOS builds are build properly (and are functional).
Hopefully no breakage in other portions

### DIFF
--- a/scripts/bz.py
+++ b/scripts/bz.py
@@ -17,7 +17,7 @@ BZ_ISDARWIN = BZ_OS == 'darwin'
 BZ_ISWIN = BZ_OS == 'windows'
 BZ_ISVERBOSE = '--verbose' in sys.argv
 
-def get_arch(short_names = True, use_mac_rosetta = True):
+def get_arch(short_names = True, use_mac_rosetta = False):
     arch = platform.machine().lower()
 
     if BZ_ISDARWIN and arch == 'arm64' and use_mac_rosetta:
@@ -28,10 +28,10 @@ def get_arch(short_names = True, use_mac_rosetta = True):
 
     if short_names and (arch in ['aarch64']):
         arch = 'arm64'
-    
+
     if short_names and (arch in ['armv7l']):
         arch = 'armhf'
-    
+
     return arch
 
 def get_os_shortname():
--- a/scripts/bz.py
+++ b/scripts/bz.py
@@ -6,6 +6,12 @@ import subprocess
 import platform
 import json
 import glob
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--verbose', action='store_true')
+parser.add_argument('--target_arch', choices=['x64', 'arm64', 'armhf'])
+args = parser.parse_args()
 
 C = {}
 
@@ -15,13 +21,9 @@ BZ_OS = platform.system().lower()
 BZ_ISLINUX = BZ_OS == 'linux'
 BZ_ISDARWIN = BZ_OS == 'darwin'
 BZ_ISWIN = BZ_OS == 'windows'
-BZ_ISVERBOSE = '--verbose' in sys.argv
 
-def get_arch(short_names = True, use_mac_rosetta = False):
+def get_arch(short_names = True):
     arch = platform.machine().lower()
-
-    if BZ_ISDARWIN and arch == 'arm64' and use_mac_rosetta:
-        arch = 'x86_64'
 
     if short_names and (arch in ['x86_64', 'amd64']):
         arch = 'x64'
@@ -33,6 +35,11 @@ def get_arch(short_names = True, use_mac_rosetta = False):
         arch = 'armhf'
 
     return arch
+    
+def get_target_arch():
+    if args.target_arch != None:
+        return args.target_arch
+    return get_arch()
 
 def get_os_shortname():
     osnames = {'linux': 'linux', 'windows': 'win', 'darwin': 'mac'}
@@ -51,6 +58,7 @@ def apply_template_vars(text):
         .replace('${BZ_VERSION}', C['version']) \
         .replace('${BZ_ARCH}', get_arch()) \
         .replace('${BZ_ARCHL}', get_arch(short_names = False)) \
+        .replace('${BZ_TARGET_ARCH}', get_target_arch()) \
         .replace('${BZ_COMMIT}', get_latest_commit())
 
 def get_target_name():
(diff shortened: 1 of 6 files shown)


scripts/bz.py
```diff
@@ -90,7 +98,7 @@ def configure_vs_tools():
         print('ERR: Unable to find VS dev command-line tools')
         sys.exit(1)
 
-    return '"%s" -host_arch=%s -arch=%s' % (vs_devcmd_file, get_arch(), get_arch())
+    return '"%s" -host_arch=%s -arch=%s' % (vs_devcmd_file, get_arch(), get_target_arch())
```
JannikGM, line 101: I consider this a bug in the original script. I'm not sure why you define `-host_arch` and `-arch` for Windows, or why this is in this special function / this particular spot in the CLI.

I'd suggest to move this to where the macOS cross-compile is added, so you can also build arm64 builds for Windows (assuming there's a cross compiler and necessary libs in VS, as there is on macOS/Xcode).

shalithasuranga: Hey @JannikGM, Thanks for the suggestions. This part was built based on this [script file (Webview's fork)](https://github.com/neutralinojs/webview/blob/3f9742ac25445e183c9807a1feb2984ee97e3a87/script/build.bat#L43) and I think it's used to configure MSVC. Perhaps we can test this section further when we try to generate ARM64 binaries on Windows?

scripts/bz.py
```diff
@@ -6,6 +6,12 @@
 import platform
 import json
 import glob
+import argparse
+
+parser = argparse.ArgumentParser()
```
shalithasuranga, line 11: `argparse` looks nice when we use several CLI flags seems, Thanks for using it :tada:

scripts/bz.py
```diff
@@ -15,14 +21,10 @@
 BZ_ISLINUX = BZ_OS == 'linux'
 BZ_ISDARWIN = BZ_OS == 'darwin'
 BZ_ISWIN = BZ_OS == 'windows'
-BZ_ISVERBOSE = '--verbose' in sys.argv
 
-def get_arch(short_names = True, use_mac_rosetta = False):
+def get_arch(short_names = True):
     arch = platform.machine().lower()
 
-    if BZ_ISDARWIN and arch == 'arm64' and use_mac_rosetta:
```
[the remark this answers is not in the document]
shalithasuranga: Hope you've mentioned line 23. Ah, initially we used that line to generate a `x64` binary on `arm64` systems since we didn't generate a specific `arm64` binary

shalithasuranga (approved):
Thanks for the contribution :tada: Would you like to display your GitHub profile name (with a link) at https://neutralino.js.org/docs/contributing/committers? -- We often update that list with developers who did great contributions :)

scripts/bz.py
```diff
@@ -215,11 +215,11 @@ def build_compiler_cmd():
         elif target_arch == 'arm64':
             cmd += '-arch arm64 '
         else:
-            print('ERR: Unsupported target architecture for macOS cross-compile: %s' % (target_arch))
+            print('ERR: Unsupported target architecture for macOS cross-compile: %s' % target_arch)
             sys.exit(1)
     else:
         if target_arch != arch:
-            print('ERR: Unable to cross-compile: target %s host %s' % (target_arch, arch))
+            print('ERR: Unable to cross-compile: target %s host %s' % target_arch, arch)
```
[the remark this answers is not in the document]
shalithasuranga: ah.. Thanks for the finding. I will correct it :)

Issue #1023, opened 2022-11-06
The problem of Chinese garbled code and crash in the configuration files and exe file names of the program


qilecms:
First question
：Use Chinese exe to pack blank or crash
![image](https://user-images.githubusercontent.com/13361784/200163149-0dfd88e6-c594-41f6-a57d-236d1a90cf0f.png)

Second question
![image](https://user-images.githubusercontent.com/13361784/200163172-1c845bbc-6a3e-41f3-8a5e-e6fe8b32bcf5.png)
![image](https://user-images.githubusercontent.com/13361784/200163211-27102668-a206-484c-a627-7b02886af4a6.png)

PR #1026, opened 2022-11-10
 The problem of Chinese garbled code and crash in the configura…

Durgesh4993:
…tion files and exe file names of the program

The problem of Chinese garbled code and crash in the configuration files and exe file names of the program solution in c++ language

## Changes proposed

 - Change 1
 - Change 2
 - etc..

## How to test it

 - Run specs/tests

## Next steps

None.

## Deploy notes

None.


JannikGM:
I assume this is meant to close https://github.com/neutralinojs/neutralinojs/issues/1023, but it just creates a garbage example file which shows an API that might be useful.

Issue #1031, opened 2022-11-18
"Move to trash"-functionality


stephanbogner:
**Is your feature request related to a problem? Please describe.**
Instead of deleting a file using [filesystem.removeFile(filename)](https://neutralino.js.org/docs/api/filesystem/#filesystemremovefilefilename) I'd like to move them to the trash. (Basically Electron's [trashItem](https://www.electronjs.org/docs/latest/api/shell#shelltrashitempath))
This way users have to option to retrieve them in case of accidental removal.

**Describe the solution you'd like**
`filesystem.trashFile(filename)` analogous to removeFile.

**Describe alternatives you've considered**
I tried to use `filesystem.moveFile(source, destination)`, but [`os.getPath(name)`](https://neutralino.js.org/docs/api/os/#osgetpathname) doesn't have `trash` as a parameter, so I don't really know any workaround.

stephanbogner:
**Some investigation:**
- The [copyPath-function is located os.cpp file](https://github.com/neutralinojs/neutralinojs/blob/7044e0f9ed326f760e7f2851ac53befd6c697e83/api/os/os.cpp#L180)
- The functionality itself is located in platform_folders.cpp, e.g. [getPicturesFolder()](https://github.com/neutralinojs/neutralinojs/blob/332e7d4915081643b32dcf870d93d97bea2dc09d/lib/platformfolders/platform_folders.cpp#L341)
- The Trash locations on different OSes to enhance `os.getPath("trash")`: 
	- [On Mac apparently](https://forums.macrumors.com/threads/trash-directory-location-in-hd.274745/post-3300983): `Macintosh HD/Users/{username}/.Trash`
	- [On Windows apparently](https://stackoverflow.com/a/22891433/3763660) `C:\$Recycle.Bin`
	- [Location on Linux](https://askubuntu.com/a/102106) (at least Ubuntu): `/home/{username}/.local/share/Trash`

PR #1052, opened 2022-12-26
GITHUB ISSUE #1031. Added trash path. os.getPath('trash')

ashu26jha:
## To use the API
     Neutralino.filesystem.movetoTrash('path-of-file-to-be-deleted');
## Notes:

 - Added movetoTrash in filesystem.cpp
 - With the help of moveFile(), set the destination of Trash by checking the operating system.
 - Added the it in neutralino.js.map also

Accidentally deleted the previous one 🫣


api/filesystem/filesystem.cpp
```diff
@@ -630,6 +630,41 @@ json moveFile(const json &input) {
...
+    #if defined(__linux__) || defined(__APPLE__) || defined(__FreeBSD__)
+    os::CommandResult getUserDetails = os::execCommand("whoami");
+    username = getUserDetails.stdOut;
+    username.pop_back();
+    string destination = "/Users/"+username+"/.Trash";
+    #if defined(__linux__)
+    destination = "/home/"+username+"/.local/share/Trash";
+    #endif
+    string command = "mv \"" + source + "\" \"" + destination + "\"";
+    os::CommandResult commandResult = os::execCommand(command);
```
BeyondMagic, line 651: Should check if folder actually exists, and exist if not.

On Linux that folder will only exist if the software [trash-cli](https://github.com/andreafrancia/trash-cli/) is installed, which is not in many distributions because it requires Python3 and its tree of dependencies.

BeyondMagic (changes requested):

api/filesystem/filesystem.cpp
ashu26jha: I guess I add a check for macOS as well as linux by using following command
`if test -d /destination_path; then echo "exist"; fi`
And throw an error message if trash folder not found. Would this work??

BeyondMagic: Yep, perhaps it can be more invasive and actually create the folder (I don't think many users will be annoyed by it, quite the opposite), if not, give a warning as well.

ashu26jha: Should I create a folder in NEU_CWD names as TrashNeu

BeyondMagic: Let collaborators decide, but I believe you should add the changes into this one.

ashu26jha: @shalithasuranga @BeyondMagic sir what should we name the folder if trash folder is not found?

shalithasuranga:
Hello @ashu26jha thanks for sending this merge request and thanks @BeyondMagic for reviewing. A small note for the implementation. I think adding a new function for this feature is too specific. We would like to keep our native API as minimal as possible. This API function will motivate us to add too specific APIs in the future. For example, a developer may request `emptyTrash()`, `restoreFromTrash()`, or even `computer.shutdown()` :smile:  in the future. 

So, I would like to try to implement this as follows first:

- Add `trash` support to [`os.getPath`](https://github.com/neutralinojs/neutralinojs/blob/68ece084decf99286c12cfafd07a368c20e93ac3/api/os/os.cpp#L180). Implementing (or requesting from the upstream) `sago::getTrashFolder` is the best.
- Let developers use `await os.getPath('trash')` with `filesystem.moveFile`.

But, there is a specific function (`shell.trashItem`) in Electron. @ashu26jha  Could you please check the difference between moving a file to the trash bin with file handling functions/commands (i.e., `mv`) and Electron's implementation? If we need a more specific implementation or `filesystem.moveFile` doesn't move a given path in all platforms (I'm not sure whether the `MoveFile` win32 function supports moving folders to the trashbin), I think `Neutralino.os.trashItem()` looks better than `Neutralino.filesystem.moveToTrash()` (See [this interesting discussion](https://github.com/electron/electron/issues/11555))

Please let me know if you have questions, Thanks :)

ashu26jha:
@shalithasuranga I just went Electron repository what they have done in Linux is to determine the desktop environment, for example as mentioned by @BeyondMagic it could be **`trash-cli`**, **`gvfs-trash`**. Depending upon this commands are executed

> Add trash support to [os.getPath](https://github.com/neutralinojs/neutralinojs/blob/68ece084decf99286c12cfafd07a368c20e93ac3/api/os/os.cpp#L180). Implementing (or requesting from the upstream) sago::getTrashFolder is the best.

Could you elaborate this? I want to work on this bit.

api/filesystem/filesystem.cpp
BeyondMagic: trash-cli and software alike use their own system, so invading their environment is not a just cause for us if we just move files to it, it is wrong. We either use their own commands or create a separate folder and move them into it; trash-neutralino or something alike should suffice.

BeyondMagic:
> + Let developers use await os.getPath('trash') with filesystem.moveFile.

I think this is the way to go, it's minimal and more difficult to come up with bugs in the future.

ashu26jha:
Hello maintainers and reviewers! I have implemented the **`os.getPath('trash')`**. 

Here are some screenshots:

- macOS

Neutralino MacOS

- Windows

Neutralino Windows



Note :

- One thing I need to make sure was that in Windows generally the getPath method uses **`SHGetKnownFolderPath`** (A WIN32 API) which is not applicable for virtual folders. Recycle bin is present in every drive.
- I don't have Ubuntu so I couldn't test on it. But I am attaching the references for my intuition

1.  - https://askubuntu.com/questions/102099/where-is-the-trash-folder
2. - Created a mapping of trash folder using folders `["XDG_TRASH_DIR"] = "$HOME/.local/share/Trash"; `

Issue #1053, opened 2023-01-03
Appending binary files are corrupted on Windows.


DavidRockin:
**Describe the bug**
Hello, just to start off I discovered this neat tool last night and have been hacking away. And I want to say I'm quite pleased with this so far, as it's allowed me to develop something pretty quickly. So, thank you VERY MUCH for your project. I'm the lead developer for a game, and I'm developing a launcher for the game. When it requests the manifest from our server, the launcher is able to figure out which platform-build it can use, and the manifest contains the file size and a SHA1 hash to verify things downloaded correctly. This lead me to notice appending binary data corrupts the files on Windows.

I've tested this on Linux, and it works correctly. As the file downloads, the binary data is appended correctly, and the file is valid (same size, same SHA1 checksum and can be opened.)

However on Windows, the file sizes will increase slightly, and thus corrupting the entire file, which appears to be caused by the appending fs methods not handling binary correctly? Likely due to line encodings? (see: #619)

**To Reproduce**
I've managed to write a little sample script to generate a temporary buffer, and write the entire buffer to a file, and then a test case of it appending chunks of the buffer to a file. The written buffer works well as expected, however when appending binary data is when the file gets corrupted.

```js
Neutralino.init();

async function sha1Sum(buf) {
    return await crypto.subtle.digest('sha-1', buf)
        .then(a => Array.from(new Uint8Array(a))
        .map(z  => z.toString(16).padStart(2,'0')).join(''));
}

Neutralino.events.on('ready', async () => {
    let buff = new ArrayBuffer(1024 * 1024);
    let view = new Uint8Array(buff);

    for (let i = 0; i < buff.byteLength; ++i) {
        view[i] = i % 256;
    }

    let i = 0;
    const fn = 'test.bin';
    console.log('filename = ' + fn + ', sha1 = ', await sha1Sum(buff));

    await Neutralino.filesystem.writeBinaryFile('good_'+fn, buff);

    while (i < buff.byteLength) {
        const chunk = buff.slice(i, i + 500);
        console.log('chunk = ', chunk.byteLength);
        await Neutralino.filesystem.appendBinaryFile('corrupt_' + fn, chunk);
        i += chunk.byteLength;
    }
});
```

**Expected behavior**
The generated `good_test.bin` is ok, and the `corrupt_test.bin` should be **EXACTLY** the same as the previous file, with the same file size and same exact SHA1 hash.

**Screenshots/Screen recordings**
[...]

PR #1054, opened 2023-01-03
[#1053] Fix Windows Binary Corruption

DavidRockin:
## Description
See #1053 for more details or sample test code. But this _should_ fix binary data being appended getting corrupted (due to Window EOL.) But this seems to fix my issues where downloaded zip files are corrupted on Windows. :)

## Changes proposed
- Handle  appending data as binary as well

## How to test it

 - Run specs/tests
 - Can see the sample code in #1053

## Next steps

None.

## Deploy notes
None.

### DIFF
--- a/api/filesystem/filesystem.cpp
+++ b/api/filesystem/filesystem.cpp
@@ -167,7 +167,7 @@ bool writeFile(const fs::FileWriterOptions &fileWriterOptions) {
     ios_base::openmode mode = ios_base::out | ios_base::binary;
     // For portability we use LF ('\n') always via ::binary
     if(fileWriterOptions.append) {
-        mode = ios_base::app | ios_base::binary;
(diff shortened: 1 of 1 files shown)


shalithasuranga:
Hey @DavidRockin  thanks for the fix, this looks great. Btw do you think it's better to avoid using ios_base::binary twice as follows:

```c++
  ios_base::openmode mode = ios_base::out | ios_base::binary;
  // For portability we use LF ('\n') always via ::binary
  if(fileWriterOptions.append) {
      mode |= ios_base::app; // Note "|"
  }
```

Note: I haven't tested it, hopefully, it works?

Thanks

shalithasuranga (approved):
I've updated it as suggested. Thanks so much for your contribution :tada:

PR #1061, opened 2023-02-05
Fixed the popup issue in windows.

kodaloid:
## Update Notes

This was originally implemented in PR #950 which has now been closed. I've updated it to work with the latest working version of the repo.

## Description
On Windows if you launch Neutralino with `hidden: true` in the config. A window appears for a split second before disappearing. Quite a few people have expressed a wish to have this changed so that the window never appears with this setting, so this pull request introduces a solution that should work given the conditions.

I tested a lot of solutions before hand. Ideally using SW_HIDE when the window is first created would be the solution, however if you do that, the WebView component always fails to load, with a corrupt connection to it's developer tools.

## Changes proposed
The changes made in this pull request are all scoped specifically for Windows using `#if defined(_WIN32)` when needed:

- Set the X coordinate to a crazy high number when CreateWindow is called.
- Change the ex window style to WS_EX_TOOLWINDOW so that no icon appears in the taskbar.
- Provide two helper methods called `window::isFakeHidden()` and  `window::undoFakeHidden()`.

If the window didn't need to hide, or show() is called, the helper methods above provide a way to show the window again to the user.

## How to test it
Build and use the new binary, everything should work as normal, however when you set hidden to true, there will no longer be a window flashing before it hides.

### DIFF
--- a/api/window/window.cpp
+++ b/api/window/window.cpp
@@ -142,6 +142,27 @@ bool isVisible() {
     #endif
 }
 
+bool isFakeHidden() {
+	//ensureOnScreen
+	#if defined(_WIN32)
+	RECT rect = { NULL };
+	
+	if(GetWindowRect( windowHandle, &rect)) {
+		return rect.left > 9999;
+	}
+	#endif
+	return false;
+}
+
+void undoFakeHidden() {
+	#if defined(_WIN32)
+	ShowWindow(windowHandle, SW_HIDE);
+	SetWindowLong(windowHandle, GWL_EXSTYLE, nativeWindow->m_originalStyleEx);
+	SetWindowPos(windowHandle, NULL, 10, 10, 0, 0, 0x0004 | 0x0001);
+	ShowWindow(windowHandle, SW_SHOW);
+	#endif
+}
+
 void show() {
     if(window::isVisible())
         return;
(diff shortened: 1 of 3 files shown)


kodaloid:
I'm not sure why this failed, any chance we can re-run the checks?

kodaloid:
I've gone through the code a fair few times now, built & tested on Windows, MacOS and Linux without any issue. I can see some other PRs were merged having similar failures, so I'm going to stop trying to fix a problem that might not be the fault of the PR.

shalithasuranga:
Hey @kodaloid thanks for working on this issue. Seems like there is an issue with our test suite (I've created #1079 to investigate it). This is finally merged and we can publish these changes with the next release. Please note that I've did a few changes with [this commit](https://github.com/neutralinojs/neutralinojs/commit/e8c4fb6156d5d49ebac0cd4a5154b1972d3aaba6) after merging

Thanks for the contribution :tada:

kodaloid:
That's great news, thanks for giving this your attention. With this sorted I can go have a poke around other issues now to see if I can contribute in other ways. Thanks again :)

shalithasuranga:
Hey @kodaloid Thanks so much, When you start contributing again, could you please check #1080? Seems like the previous version of the code also have this issue

Thanks :tada:

PR #1064, opened 2023-02-08
Add net API

manavagr1108:
Roadmap feature
## Description

- Added networking API called `net`
- Uses `curl` to fetch the data from the requested url
- Returns the JSON response in the string format

## How to test it

- Tested it by building the client


manavagr1108:
Hey @shalithasuranga Can please have a look at this pr and let me know if this is good to go or not? It would be really helpful, thank you!

shalithasuranga:
Hey.. Thanks for sending this PR. As I remember, I also tried to use CURL and some C++ wrappers to implement the `net` module. The issue was CURL is not embedded as an inbuilt OS package. Here we can see the issue in automated builds. Would you like to research this issue further? Maybe we can try https://github.com/yhirose/cpp-httplib's client? (There are also some extra dependencies it seems)

Thanks :tada:

shalithasuranga:
Hey.. I am closing this. Feel free to discuss and re-open if needed. Thanks :tada:

IsmaCortGtz:
Hi! I'm very interested in this feature. I was reading about [cpp-httplib](https://github.com/yhirose/cpp-httplib) and I have some ideas.

Itself, this library only allows HTTP requests; for HTTPS we need to include OpenSSL, but this need to install a different dev dependencie for each platform, and it could even not be installed in some users' devices.

The other option I found was [cpp-httplib](https://github.com/yhirose/cpp-httplib) + [mbedtls](https://github.com/Mbed-TLS/mbedtls), but it could be slower for binary data like large videos, images, and streaming.

I think [mbedtls](https://github.com/Mbed-TLS/mbedtls) is the best option for almost all cases, but it would be worse than some users could expect.

@shalithasuranga, What do you think is the best option?

IsmaCortGtz:
I compiled a new static version using `mbedtls` instead of `openssl` and the results are very good.

Screenshot_20250616_100502

The size of the three files is after `stripe`, without it, all should be around `900 KB` bigger. This build should work in every Linux regardless of installed software. But this means only a `1 MB` increase in the current `neutralinojs` build.

Issue #1058, opened 2023-01-13
`http://localhost:{NL_PORT}/neutralino.js` returns 404 but it works


Vexcited:
**Describe the bug**
When going to `http://localhost:{NL_PORT}/neutralino.js`, it gives 404 Not Found in the headers while the content body is here.

**To Reproduce**
`curl.exe -L -i http://localhost:4000/neutralino.js` (here, my `{NL_PORT}` is `4000`)

It returns:

```
HTTP/1.1 404 Not Found
Content-Length: 557
Content-Type: application/javascript
Server: WebSocket++/0.8.2

var NL_OS='Windows';var NL_VERSION='4.9.0';var NL_COMMIT='766c5e0debe0d3b2534e5662b90a7a6a66e0693b';var NL_APPID='com.github.Vexcited.SomeRandomProject';var NL_APPVERSION='0.0.0';var NL_PORT=4000;var NL_MODE='cloud';var NL_TOKEN='';var NL_CWD='C:/Users/user/path/to/project';var NL_ARGS=["bin\\neutralino-win_x64.exe","","--load-dir-res","--path=.","--export-auth-info","--neu-dev-extension","--neu-dev-auto-reload","--port=4000","--mode=cloud"];var NL_PATH='.';var NL_PID=880;var NL_RESMODE='directory';var NL_EXTENABLED=false;var NL_CMETHODS=[];  
```

**Expected behavior**
Have a 200 OK as response in the header.

**Specifications**  
 - OS: Windows 10 Pro 21H2
 - Neutralinojs version: v4.9.0
 - Neutralinojs client library version: v3.8.0
 - Neutralinojs CLI version: v9.3.1

parteekcoder:
hey @Vexcited , would you like to assign this issue to me?

Vexcited:
Sorry @parteekcoder but I can't do that, I don't have the rights to edit the assignees of my issue-

shalithasuranga:
Fixed via #1076 Thanks for reporting and discussing :tada:

PR #1076, opened 2023-02-26
Fix: Invalid file path in HTTP request returns global variables

MrVSiK:
## Description
Currently, a HTTP 404 status code is returned when the file path is invalid, however the message body contains all the global variables instead of a valid error message (Refer to issue #1058 ). With this PR, the router will now check if the file is read successfully before adding content to the response data.

## Changes proposed

 - Check for `fileReaderResult.status` before checking the contents of  `prependData` 

## Next steps

- A suitable message in the response body would be beneficial but I am not sure about the content.


shalithasuranga (approved):
Thanks for the contribution :tada: @MrVSiK

Issue #960, opened 2022-07-20
No Unicode support in Windows


kolbasa:
**Describe the bug**
I am trying to build an open source application for all platforms and all languages.
In my use case, I need simple access to the file system.
Unfortunately, this is impossible for most languages on Windows.

For example, the following command cannot be executed on most computers:
```js
await Neutralino.filesystem.readDirectory('C:\\Users\\Michael'); // never resolves
```
This is because Windows places appropriately translated default folders in the home directory.

You might think that you could build a workaround with `Neutralino.os.execCommand`. 
Unfortunately, the problem runs through the entire native api.

```js
await Neutralino.os.execCommand('dir C:\\Users\\Michael'); // never resolves
```

That's a shame when you have to categorically exclude most European countries from using it.

**Specifications**  
- OS: Windows 11 x64, Windows 10 x64
- Neutralinojs version: v4.6.0
- Neutralinojs client library version: v3.5.0
- Neutralinojs CLI version: v9.3.1

pathange-s:
Hey @kolbasa, sorry for the inconveniences. There have been several efforts put in by the organisation to bring support for the same.
It was also included in the GSoC ideas list to get the support asap. Check [here](https://github.com/neutralinojs/gsoc2022#2-support-unicode-characters-in-neutralinojs-windows-version). But, still developers are working on it to get a workaround for this.

kolbasa:
Not a problem. Neutralinojs is a great project! Eventually it will work! 👍

onemen:
Can anyone update when this will be fixed?

kolbasa:
I have given up hope. The problem has existed for so long. I would switch to another framework.

pegvin:
you can look at pegvin/arujs but it's very new tho but is Unicode compatible.

onemen:
@pegvin
> you can look at pegvin/arujs but it's very new tho but is Unicode compatible.

Is it possible to use some of the code from arujs as an extension in neutralino to replace `Neutralino.filesystem` ?

pegvin:
@onemen nopes, cause AruJs Uses The IPC Provided By WebView Itself and Neutralino Uses WebSockets To Transfer All The Data...

onemen:
@pegvin
is it possible to compile AruJs code to an exe, then i can call it with   `Neutralino.os.execCommand(arujs_code.exe);`
i do it now for an hash function that i use, without an extension.

pegvin:
[...]

Issue #808, opened 2022-02-22
starter template crashes on windows 10 - error code 3221225477


gui3:
**Expected Behavior**  
I tried running the starter template following the instructions on [this link](https://neutralino.js.org/docs/getting-started/your-first-neutralinojs-app) on my work machine.
- neu is istalled globally.
- **I tried running the `CheckNetIsolation.exe` command before -> no changes**
- **i tried installing WebView2 runtime -> no changes**
  
**Actual Behavior**  
the application crashes instantly after `neu run`, 
the app window is blank and this dialog appears :

![erreur_neutralino_2](https://user-images.githubusercontent.com/45894433/155148744-4186e690-8c26-4ce9-9138-4916023aadb4.jpg)

the french message says "Microsoft Edge cannot read nor write in his data directory : "

The console shows the following : 
```
$ neu run
neu: INFO Starting process: neutralino-win_x64.exe  --load-dir-res --path=. --export-auth-info --neu-dev-extension --neu-dev-auto-reload
INFO 2022-02-22 13:57:16,871 Auth info was exported to ./.tmp/auth_info.json api/debug/debug.cpp:14 <my-user-name>@AIB-A0064-P
neu: INFO neu CLI connected with the application.
neu: WARNING neutralino-win_x64.exe was stopped with error code 3221225477
```

Since it's a work machine, I have **no control over Windows updates** and my version (windows 10 version 21H1 build 19043.1083) dates back to may 2021. This could be a source of the problem ?

**Steps to Reproduce the Problem**  
 1. get a similar windows build?
 2. `npm install -g @neutralinojs/neu`
 3. `neu create <project name>` project name was 'test' in my case
 4. `cd <project name>`
 5. `neu run`

**Specifications**  
* NeutralinoJs Version: binaries 4.3.0 - client 3.2.0
* NeutralinoJs CLI version: 9.2.0
* Platform: windows 10 version 21H1 build 19043.1083

tukusejssirs:
It crashes on Arch Linux, too.

```bash
--- Global ---
neu CLI: v9.3.1

--- Project: ---
Neutralinojs binaries: v4.6.0
Neutralinojs client: v3.5.0
```

pathange-s:
@gui3 Please check if this issue is because of the reasons mentioned in [here](https://github.com/neutralinojs/neutralinojs/issues/613). If yes, there's a temporary fix that can help you out. We are working on bringing support for Unicode, and hope it will be out soon.

[...]

Issue #782, opened 2022-01-30
filesystem.readDirectory fails when results would return filenames containing umlauts


sroauth:
**Expected Behavior**
Neutralino's native filesystem API should return a list of files when using the `readDirectory` method.
  
**Actual Behavior**
Neutralino is not returning anything when calling the `readDirectory` method.

**Steps to Reproduce the Problem**
 1. Have a file named `björn-kjellström.txt` in the directory being read from.
 2. Call the `readDirectory` method on the directory containing the file `björn-kjellström.txt`.

**Specifications**
* NeutralinoJs Version: 4.2.0
* NeutralinoJs CLI version: 3.1.0
* Platform: Windows 11 Pro 21H2

DEVLOPRR:
can you share the code too?

rince1984:
The problem occurs with all filesystem operations (e.g. ```copyFile```). I also need a solution for that.

pathange-s:
@rince1984 what error did you get when you tried to run any of the `filesystem` operations ?

I tried to call this function

```
function readDir() {
    Neutralino.filesystem.readDirectory('./')
    .then((result) => console.log(result))
    .catch((err) => console.log(err));
}
```
and obtained this as the output
```
[Log] Array (11) (main.js, line 20)
0 {entry: ".", type: "DIRECTORY"}
1 {entry: ".gitignore", type: "FILE"}
2 {entry: "LICENSE", type: "FILE"}
3 {entry: "bin", type: "DIRECTORY"}
4 {entry: ".tmp", type: "DIRECTORY"}
5 {entry: "neutralinojs.log", type: "FILE"}
6 {entry: "README.md", type: "FILE"}
7 {entry: "..", type: "DIRECTORY"}
8 {entry: "resources", type: "DIRECTORY"}
9 {entry: ".github", type: "DIRECTORY"}
10 {entry: "neutralino.config.json", type: "FILE"}
```
ie It worked as expected for me.


PS : If the same function was written as 

```
function readDir() {
    let entries = await Neutralino.filesystem.readDirectory(NL_PATH);
    console.log('Content: ', entries);
}
```

It threw these errors in the console.

```
[Error] SyntaxError: Unexpected identifier 'Neutralino'. Expected ';' after variable declaration.
	(anonymous function) (main.js:19)
[Error] ReferenceError: Can't find variable: createDir
	onclick (localhost:14)
```

[...]

Issue #613, opened 2021-09-24
UTF-8 not supported in Native API


ghost:
**Expected Behavior**  
Support cyrillic encoding in window title and filesystem.
  
**Actual Behavior**
Window title is not displayed correctly. Created directory is also corrupted.
![image](https://user-images.githubusercontent.com/55308117/134721940-2588bb6a-3f55-48a2-ab9e-260eaea62cb8.png)

**Steps to Reproduce the Problem**  
 1. Set `modes.window.title` to `Пример`
 2. Change `main.js` content to 
 ```
window.myApp = {
    showInfo: () => {
        Neutralino.filesystem.createDirectory({
            path: './папка',
        });
    },
    onWindowClose: () => {
        Neutralino.app.exit();
    }
};

Neutralino.init();
Neutralino.events.on("windowClose", myApp.onWindowClose);
window.myApp.showInfo();
```
 3. `neu run`

**Specifications**  
* Neutralinojs binaries: v2.8.0
* Neutralinojs client: v1.5.0
* neu CLI: v6.2.0
* Platform: Windows 10

LTVA1:
I have this problem too.

khushit-shah:
I was not able to reproduce it on windows 10 21H2

![image](https://user-images.githubusercontent.com/42430171/161367600-1e2dc989-6ec7-418b-a7f1-57eeba2f8700.png)

But the folder was not created, maybe I am missing some permissions in the config file.

ebukaodini:
I intend to work on this issue in the [GSoC 2022 Program](https://github.com/neutralinojs/gsoc2022#2-support-unicode-characters-in-neutralinojs-windows-version).

I'd appreciate every form of support in terms of screenshots of the bug.
Anything at all that would help me replicate and fix the bug.

Thanks.

duzc2:
i have this too
--- Global ---
neu CLI: v9.3.0

--- Project: OpenBlock (js.neutralino.sample) ---
Neutralinojs binaries: v4.5.0
Neutralinojs client: v3.4.0
Project version: v1.0.0

pegvin:
Sorry but unicode isn't currently supported, but is being worked on.

duzc2:
also does not support utf-8 in request url:

http://localhost:63507/project/assets/%E5%8F%A3%E5%8F%B7.png

pegvin:
yes we know that, sorry for the inconvenience.

duzc2:
OMG
[...]

PR #1102, opened 2023-05-04
Fix Unicode in the Windows version of Neutralino.js

FelisDiligens:
## Description

Under Windows, Neutralino doesn't support Unicode, see the linked issues below.  
This pull request should hopefully fix this, by using the Unicode WinApi.

### Related issues and posts
- [GSOC2023 - Support Unicode characters in Neutralinojs Windows version](https://github.com/neutralinojs/gsoc2023#2-support-unicode-characters-in-neutralinojs-windows-version)
- [Issue #613 - UTF-8 not supported in Native API](https://github.com/neutralinojs/neutralinojs/issues/613)
- [Issue #782 - filesystem.readDirectory fails when results would return filenames containing umlauts](https://github.com/neutralinojs/neutralinojs/issues/782)
- [Issue #808 - starter template crashes on windows 10 - error code 3221225477](https://github.com/neutralinojs/neutralinojs/issues/808)
- [Issue #960 - No Unicode support in Windows](https://github.com/neutralinojs/neutralinojs/issues/960)

### Broken functions / issues
- Native API fails when working with Unicode
  - `Neutralino.filesystem.readDirectory`
  - `Neutralino.filesystem.createDirectory`
  - `Neutralino.os.execCommand`
  - `Neutralino.os.spawnProcess`
  - `Neutralino.os.updateSpawnedProcess`
  - `Neutralino.window.setTitle`
  - (possibly more)
- Neutralino crashes on start if it's in a path with e.g. "Umlaut" characters (`ä`, `ö`, `ü`, or other non-ANSI characters)
- Neutralino does not display Unicode characters in the window title correctly (non-ANSI characters are garbled or question marks, see screenshots)

#### Before
neutralino-win_x64 bak_CwgBMmDGpL

#### After
neutralino-win_x64_zKsWTHBbQO

## Changes proposed
- Define `UNICODE` and `_UNICODE` at compile time to force the use of wide string versions of the WinApi.
- Use `wWinMain` and `__wargv` instead of `WinMain` and `__argv`.
- Replace -`A` WinApi calls with generic calls: e.g. `PathRemoveFileSpecA` → `PathRemoveFileSpec` (will expand to `PathRemoveFileSpecW`)
- Convert between `std::string` and `std::wstring`, as well as `char*` and `wchar_t*` where needed.
- Use `L""` strings where needed. (or using `_T()` macro from `tchar.h`)
- Convert paths from `std::string` (UTF-8 encoding) to `std::wstring` (UTF-16 encoding) when opening an `ifstream` or an `ofstream`. (under Windows, `ifstream` and `ofstream` only support ANSI or UTF-16)

### Sources
<details>
<summary>Click to see sources</summary>

- [WinAPI: A and W functions](https://renenyffenegger.ch/notes/Windows/development/WinAPI/A-and-W-functions)
- [TEXT vs. _TEXT vs. _T, and UNICODE vs. _UNICODE](https://devblogs.microsoft.com/oldnewthing/20040212-00/?p=40643)
- [Read and write binary files with UTF-8 file names on windows](https://stackoverflow.com/a/67577383)
- [Conversion methods](https://github.com/samhocevar/portable-file-dialogs/blob/67e7b0945aac80efa2ec5c72de98b47f7552735c/portable-file-dialogs.h#L420)
- [FAQ - JSON for Modern C++ - Parse errors reading non-ASCII characters](https://json.nlohmann.me/home/faq/#parse-errors-reading-non-ascii-characters)
- [UTF-8 Everywhere - 10. How to do text on Windows](https://utf8everywhere.org/#windows)
- [MSDN - Generic-Text Mappings in tchar.h](https://learn.microsoft.com/en-us/cpp/text/generic-text-mappings-in-tchar-h?view=msvc-170)

</details>

## Notes

I tried my best to test this PR (including running and passing the test suite), but I'm not sure, since it changes a lot.
It compiles and runs on Windows and GNU/Linux, but I couldn't test macOS, since I don't own a mac.  
Also, I added a `utils` folder, since I didn't know where to best put `str_conv.cpp`.

### DIFF
--- a/api/filesystem/filesystem.cpp
+++ b/api/filesystem/filesystem.cpp
@@ -17,6 +17,8 @@
 #include <shlwapi.h>
 #include <winbase.h>
 
+#include "utils/win/str_conv.cpp"
+
 #define NEU_WINDOWS_TICK 10000000
 #define NEU_SEC_TO_UNIX_EPOCH 11644473600LL
 #endif
@@ -120,7 +122,7 @@ bool createDirectory(const string &path) {
     #if defined(__linux__) || defined(__APPLE__) || defined(__FreeBSD__)
     return mkdir(path.c_str(), 0700) == 0;
     #elif defined(_WIN32)
-    return CreateDirectory(path.c_str(), nullptr) == 1;
+    return CreateDirectory(str2wstr(path).c_str(), nullptr) == 1;
     #endif
 }
 
@@ -128,7 +130,7 @@ bool removeFile(const string &filename) {
     #if defined(__linux__) || defined(__APPLE__) || defined(__FreeBSD__)
     return remove(filename.c_str()) == 0;
     #elif defined(_WIN32)
-    return DeleteFile(filename.c_str()) == 1;
+    return DeleteFile(str2wstr(filename).c_str()) == 1;
     #endif
 }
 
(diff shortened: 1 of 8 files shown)


shalithasuranga (approved):
Thanks so much for your contribution. This solves a critical, long-waited issue in the framework! :tada:

PR #1123, opened 2023-06-06
Restore Win32 window style before showing it (to avoid toolwindow style)

nicorac:
## Description
I'd like my Win32 (window mode) application to restore its last native window position/size at startup **without glitches**.

This is how I've implemented it:

- `neutralino.config.json` with `modes.window.hidden` set to `true`
- web application (Angular) is bootstrapped
- last window position is loaded (from own config file)
- native window is moved/resized with `Neutralino.window.move()` and `Neutralino.window.setSize()`
- window is shown with `Neutralino.window.show()`

## Current behavior

The native window appears with the (wrong) toolwindow style.

## Expected behavior

Window should appear with the default style (min, max, close buttons).

## Changes proposed

This is caused by the window never being shown at least once before move/size calls.
It could be easily fixed by restoring original window style just before showing it.

## Additional change

I've also uniformed `move()` function to use `SetWindowPos()` instead of `GetWindowRect() + MoveWindow()` as in other similar calls.

### DIFF
--- a/lib/webview/webview.h
+++ b/lib/webview/webview.h
@@ -1336,10 +1336,25 @@ public:
   }
   void *window() { return (void *)m_window; }
   void terminate(int exitCode = 0) {
+
+    // event to wait for window close completion
+    auto evtWindowClosed = CreateEvent(
+        NULL,               // default security attributes
+        TRUE,               // manual-reset event
+        FALSE,              // initial state is nonsignaled
+        TEXT("WindowClosedEvent")  // object name
+    );
     processExitCode = exitCode;
+
     dispatch([=]() {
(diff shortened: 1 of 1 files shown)


nicorac:
Well, after a deep test I've seen that this fix is only partial: it still does not work if the window is maximized just before calling  `show()`.

I think the base issue is related to `WS_EX_TOOLWINDOW` trick, used to avoid native window to be "hosted" in systray.

A better fix should be the one posted here on SO: https://stackoverflow.com/questions/30933219/hide-window-from-taskbar-without-using-ws-ex-toolwindow

nicorac:
Will create a new one, closing this

PR #1127, opened 2023-06-14
Improved Win32 window style handling with ~WS_EX_APPWINDOW

nicorac:
## Description
`WS_EX_TOOLWINDOW` is not the best way to remove taskbar icon for an application, because it has the counter-effect to change the visual style of the affected window.

## How to reproduce
My application window starts hidden (thanks to `window-hidden: true` in `neutralino.config.json`.
Just after start it reads last window position/size, then calls `Neutralino.window.show()` to show the window.
Doing this, the native window is missing the minimize/maximize buttons and it appears as a tool-window (with only the small close button).

## Changes proposed
Best way to remove icon is to create a window **without** the `WS_EX_APPWINDOW` style, and then ensure it's focused each time it's shown/hidden/maximized/restored.

### DIFF
--- a/api/window/window.cpp
+++ b/api/window/window.cpp
@@ -184,6 +184,7 @@ void maximize() {
     gtk_window_maximize(GTK_WINDOW(windowHandle));
     #elif defined(_WIN32)
     ShowWindow(windowHandle, SW_MAXIMIZE);
+    SetForegroundWindow(windowHandle);
     #elif defined(__APPLE__)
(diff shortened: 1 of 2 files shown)


nicorac:
Can't explain why these tests are failing; unrelated changes...

shalithasuranga:
@nicorac, Please do worry about test failures, I will test these on the Windows VM and merge them soon :tada: The changes looked great, I just wanted to test binaries before merging ;)

shalithasuranga (approved):
Thanks so much for your contribution :tada: This looks great :)

PR #1199, opened 2024-02-13
test runner uses process.arch on mac

monsterkodi:
## Description
test runner uses process.arch on mac, like on linux

## How to test it
Run specs/tests on an arm based mac


shalithasuranga (approved):
Thanks so much for this contribution @monsterkodi :tada:

PR #1219, opened 2024-03-08
Added Support For Notification using WinToast for Windows 11

NayanPatidar:
## Description
   This PR Provides support for the Notification in Windows 11 using WinToast Library. It is necessary as pfd (Portable File Dialog ) does not provide support for the Windows 11

## Changes proposed
 - Added WinToast library
 - Notification Code Changed for Windows only! (For Mac and Linux still using the pfd)

## How to test it
- Use  - await Neutralino.os.showNotification('Hello world', 'It works! Have a nice day'); in the main.js

## Next steps

None.

## Deploy notes

None.


shalithasuranga:
Hello @NayanPatidar, thanks so much for sending this pull request. Before linking this new library I would like to ask some help. Could you please examine why pfd doesn't work on Win11? Is it possible to fix it on pfd without requesting something new from the upstream? If that's doable I think we can fix pdf and use it without linking another lib... Thanks :tada:

NayanPatidar:
> Hello @NayanPatidar, thanks so much for sending this pull request. Before linking this new library I would like to ask some help. Could you please examine why pfd doesn't work on Win11? Is it possible to fix it on pfd without requesting something new from the upstream? If that's doable I think we can fix pdf and use it without linking another lib... Thanks 🎉

Sure, give me some time. I will definitely figure out what is causing the issue and let you know !

shalithasuranga:
Just closing to make the PR section clean -- we can indeed re-open this later once we have new commits or dicussions :)

PR #1222, opened 2024-03-17
setupwebscoket

Parv-gugnani:
Refactored code for WebSocket communication setup, message handling, and logging with improved organization.


shalithasuranga (commented):
Hello @Parv-gugnani, We would like to keep these files as it is without these changes. I hope you'll do another contribution to Neutralinojs. Thanks so much for sending this PR :tada:

Issue #1179, opened 2023-12-04
When exitProcessOnClose is set to true, closing the app will result in +5 sec freeze before the app closes on Windows


Nassim-Saboundji:
**Describe the bug**
When `exitProcessOnClose` is set to true, closing the app will result in +5 sec freeze before the app closes on Windows

**To Reproduce**
1. set `exitProcessOnClose` to true in `neutralino.config.json`.
2. Open the app.
3. Try closing it.

**Expected behavior**
The app should close immediately.

**Specifications**  
 - OS: Windows 11
 - Neutralinojs version : 4.14.1
 - Neutralinojs client library version: 3.12.0
 - Neutralinojs CLI version: 10.1.0

*Note: you can enter the `neu version` command to find these details*

**Additional context**
The issue does not occur when using this workaround in main.js instead of relying on existProcessOnClose.
```js
Neutralino.init();
Neutralino.events.on("windowClose", () => {
  Neutralino.app.exit();
});
```

ContrerasA:
Something similar happens when creating a react application as outlined under the "Using any frontend library" section of the docs

https://neutralino.js.org/docs/getting-started/using-frontend-libraries

 In my case, when I press the X button to exit, the application hangs for 10 seconds before closing.
I have not found how to use the workaround shown above since there is no main.js file

thorin8k:
Same here, tested in 4.15.0 and same results.

The workaround seems to be working fine.

Rylazius:
same, neutralino v4.14.1, client v3.12.0, workaround working fine but not the `exitProcessOnClose`

multiman95:
> **Describe the bug** When `exitProcessOnClose` is set to true, closing the app will result in +5 sec freeze before the app closes on Windows
> 
> **To Reproduce**
> 
> 1. set `exitProcessOnClose` to true in `neutralino.config.json`.
> 2. Open the app.
> 3. Try closing it.
> 
> **Expected behavior** The app should close immediately. However, we have a +5 sec delay before closing. Which is not what we want.
> 
> **Specifications**
> 
> * OS: Windows 11
> * Neutralinojs version : 4.14.1
> * Neutralinojs client library version: 3.12.0
> * Neutralinojs CLI version: 10.1.0
> 
> _Note: you can enter the `neu version` command to find these details_
> 
> **Additional context** The issue does not occur when using this workaround in main.js instead of relying on existProcessOnClose.
> 
> ```js
> Neutralino.init();
[...]

PR #1243, opened 2024-04-08
Reduce delay before app closes on windows

Spishous:
## Description
in certain Windows configurations, the signal expected by "WaitForSingleObject" before closing the application window (when "exitProcessOnClose" is true in neutralino.config.json), is not received. Which creates a freeze lasting from 2sec to 10sec on some machines. #1179 

## Changes proposed

 - Reduced WaitForSingleObject delay

## Next steps

I don't know enough about C++ to delve deeper into the problem. But the next step would be to see why when "exitProcessOnClose" is true, the behavior of the closure changes and blocks at this step.

None.

### DIFF
--- a/lib/webview/webview.h
+++ b/lib/webview/webview.h
@@ -947,7 +947,11 @@ public:
     });
 
     // wait for dispatch() to complete
-    WaitForSingleObject(evtWindowClosed, 10000);
+    /* TODO: Check why this method doesn't trigger a signal on Windows
+    *        when "exitProcessOnClose" is true. (Previous value: 10000)
+    *        This wait causes a delay when closing the program.
+    */
+    WaitForSingleObject(evtWindowClosed, 300);
     CloseHandle(evtWindowClosed);
 
   }


Spishous:
@nicorac

nicorac:
Well, the `WaitForSingleObject` delay is not really a "delay" but a timeout: the function waits for at most 10000ms for the watched handle to be released (at window destruction).
It was added because the applicaiton did not correctly release its resources, leading to handle leaks.

That said, the question is: is this new behavior related to Windows 11?
The OP of #1179 runs Win11, but others not, so I'm not sure...

Sadly I'm actually not able to further investigate on it, I'm sorry 😔

moritz-baecker-integra:
> Well, the `WaitForSingleObject` delay is not really a "delay" but a timeout: the function waits for **at most** 10000ms for the watched handle to be released (at window destruction). It was added because the applicaiton did not correctly release its resources, leading to handle leaks.
> 
> That said, the question is: is this new behavior related to Windows 11? The OP of #1179 runs Win11, but others not, so I'm not sure...
> 
> Sadly I'm actually not able to further investigate on it, I'm sorry 😔

This behavior also happens on Windows 10 (Version 10.0.19045 Build 19045). I have to wait multiple seconds for the window to close.

AntumDeluge:
Same here, on Windows 10.

shalithasuranga (approved):
Thanks so much @Spishous for sending this pull request :tada:

PR #1250, opened 2024-05-04
Add transparent windows support on Windows

### LINKED ISSUE #1201: Implement `window.transparent` configuration option for Windows
Currently, the `window.transparent` config prop is only implemented in GNU/Linux and macOS as follows:

- GNU/Linux: Window transparency is set in the [`webview.h`](https://github.com/neutralinojs/neutralinojs/blob/d8cf72c737d0ad3b598904661a63ba18c15654c2/lib/webview/webview.h#L473) file and the webview transparency is set in the [same file](https://github.com/neutralinojs/neutralinojs/blob/d8cf72c737d0ad3b598904661a63ba18c15654c2/lib/webview/webview.h#L579).
- macOS: Window transparency is set in the [`macwindow.mm`](https://github.com/neutralinojs/neutralinojs/blob/d8cf72c737d0ad3b598904661a63ba18c15654c2/lib/webview/macwindow.mm#L19) file and the webview transparency is set in the [`webview.h`](https://github.com/neutralinojs/neutralinojs/blob/d8cf72c737d0ad3b598904661a63ba18c15654c2/lib/webview/webview.h#L839) file

Similarly, we need to implement this conditional transparency feature in Windows too. 
- Window transparency: https://stackoverflow.com/questions/3970066/creating-a-transparent-window-in-c-win32
- Webview2 transparency: https://stackoverflow.com/questions/67838739/how-to-use-transparency-in-webview2

ArcticFqx:
## Description
This enables support for transparent windows on Windows.
Resolves #1201.

image

## Changes proposed
* Add WS_EX_LAYERED to GWL_EXSTYLE
* Overwrite GWL_STYLE and only keep WS_THICKFRAME, this in effect makes the window borderless, apart from the drag handles for resizing.
* Set the `WEBVIEW2_DEFAULT_BACKGROUND_COLOR` environment variable to make the WebView2 background transparent white.
* Set the initial window size to 8000x8000. There is an issue I haven't been able to figure out where the initial window size of a layered window dictates the active area of the window. This works around it for most cases, as most people don't have monitors with greater than 8k resolution.

## How to test it
Set `"transparent": true` in `neutralino.config.json`.

### DIFF
--- a/lib/webview/webview.h
+++ b/lib/webview/webview.h
@@ -881,7 +881,7 @@ public:
           });
       RegisterClassEx(&wc);
       m_window = CreateWindow(L"Neutralinojs_webview", L"", WS_OVERLAPPEDWINDOW, 99999999,
-                              CW_USEDEFAULT, 640, 480, nullptr, nullptr,
+                              CW_USEDEFAULT, 8000, 8000, nullptr, nullptr,
                               GetModuleHandle(nullptr), nullptr);
       SetWindowLongPtr(m_window, GWLP_USERDATA, (LONG_PTR)this);
     } else {
@@ -889,6 +889,15 @@ public:
     }
 
     setDpi();
+
+    if (transparent) {
+      // remove other window styles and keep only the border
+      SetWindowLongPtr(m_window, GWL_STYLE, WS_THICKFRAME);
+      SetWindowLongPtr(m_window, GWL_EXSTYLE, GetWindowLongPtr(m_window, GWL_EXSTYLE) | WS_EX_LAYERED);
+
+      // transparent white, use of environment variable prevents flashing on show
+      SetEnvironmentVariable(L"WEBVIEW2_DEFAULT_BACKGROUND_COLOR", L"00FFFFFF");
+    }
 
     // stop the taskbar icon from showing by removing WS_EX_APPWINDOW.
     SetWindowLong(m_window, GWL_EXSTYLE, GetWindowLong(m_window, GWL_EXSTYLE) & ~WS_EX_APPWINDOW);


shalithasuranga:
Hello @ArcticFqx thanks so much for sending this pull request. Please ignore CI test failures (There is an issue with the Windows test). I'll merge this soon :tada:

lib/webview/webview.h
```diff
@@ -881,7 +881,7 @@ class win32_edge_engine {
           });
       RegisterClassEx(&wc);
       m_window = CreateWindow(L"Neutralinojs_webview", L"", WS_OVERLAPPEDWINDOW, 99999999,
-                              CW_USEDEFAULT, 640, 480, nullptr, nullptr,
+                              CW_USEDEFAULT, 8000, 8000, nullptr, nullptr,
```
servusdei2018, line 884: Is it necessary to make the window so large?

Basti-def: Yes. There is something about this in the description of the Pull Request.
There was an issue otherwise.

servusdei2018: The issue may stem from creating the window as `WS_OVERLAPPEDWINDOW` see [Win32 docs](https://learn.microsoft.com/en-us/windows/win32/winmsg/window-styles).

It may be worthwhile to call `CreateWindowEx` instead like so:
```c++
m_window = CreateWindowEx(WS_EX_LAYERED, L"Neutralinojs_webview", L"", WS_OVERLAPPEDWINDOW,
                          CW_USEDEFAULT, CW_USEDEFAULT, 640, 480, nullptr, nullptr,
                          GetModuleHandle(nullptr), nullptr);
```

and then enable transparency via `SetLayeredWindowAttributes(m_window, 0, XXX, LWA_ALPHA);` where `XXX` is the alpha value.

ArcticFqx: Yeah, that does work for the entire window and it keeps the resizing behavior consistent, but this also affects the transparency of the webview content, which I think we don't want to happen as we want to control that with CSS instead.

It's a good start though, it's half of a better solution than my proposed PR. We have to be a little clever if we want to keep the window itself transparent, but not the webview.

servusdei2018: Hmm, would changing
```c++
SetEnvironmentVariable(L"WEBVIEW2_DEFAULT_BACKGROUND_COLOR", L"00FFFFFF");
```
to
```c++
SetEnvironmentVariable(L"WEBVIEW2_DEFAULT_BACKGROUND_COLOR", L"FFFFFFFF");
```
prevent the webview from becoming transparent?

Sadaf-A:
Hey! @ArcticFqx, 
could you please confirm the windows version you used for testing this? It's working differently on windows 11

bxsti-dev:
> Hey! @ArcticFqx, 
> could you please confirm the windows version you used for testing this? It's working differently on windows 11

I used Windows 11 and it worked.

Sadaf-A:
> > Hey! @ArcticFqx, 
> > could you please confirm the windows version you used for testing this? It's working differently on windows 11
> 
> I used Windows 11 and it worked.

can you share a screenshot?

bxsti-dev:
> can you share a screenshot?

@Sadaf-A  here:

image

bxsti-dev:
> Hey, thanks for sharing the screenshot are you able to move the window and close it?

Unfortunately not :(  I would have wished that too. But apparently it has to be borderless. (I guess that`s still the case?)
You have to create a custom "titlebar".

lib/webview/webview.h
ArcticFqx: Yes, the format for that string is AARRGGBB.

ArcticFqx:
> Hey! @ArcticFqx, could you please confirm the windows version you used for testing this? It's working differently on windows 11

I am using Windows 11

ArcticFqx:
Is there some other requirement that I'm missing in this PR? I don't really have the other platforms to compare too.

shalithasuranga (approved):

shalithasuranga:
Thanks @ArcticFqx  for sending this PR and thanks others for reviewing. We can include this implementation since looks like we met a platform limitation here (We can research about this later). We can add a note about the window titlebar becoming transparent to the documentation. However, app developers typically use `transparent: true` to make custom window shapes that contain custom titlebars with draggable regions I think :tada:

PR #1253, opened 2024-05-16
Testsuite - app and computer APIs

Sadaf-A:
## Description
This PR adds additional testcases for app API and computer API for better coverage

## Changes proposed
Following testcases have been added
### app API
exit()
- [x] throws an error for invalid exit codes

getConfig()
- [x] JSON object contains the right fields

broadcast()
- [x] throws an error for an invalid event name
- [x] broadcasts an event with no listeners registered
- [x] handles high load by broadcasting multiple events rapidly
- [x] triggers all registered listeners when broadcasting an event
- [x] handles errors in listeners gracefully
- [x] executes listeners in the order they were added
### computer API
getMemoryInfo()
- [x] returns consistent and valid memory info

getArch()
- [x] returns a supported architecture

getMousePosition()
- [x] returns the current mouse cursor position and it is within the screen bounds
## How to test it

 - Run specs/tests

## Next steps

None.

## Deploy notes

None.


spec/app.spec.js
```diff
@@ -36,12 +46,25 @@ describe('app.spec: app namespace tests', () => {
...
     describe('app.getConfig', () => {
-        it('returns the config as a JSON object', async () => {
+        it('JSON object contains the right fields', async () => {
             runner.run(`
                 let config = await Neutralino.app.getConfig();
                 await __close(JSON.stringify(config));
             `);
-            assert.ok(typeof JSON.parse(runner.getOutput()) == 'object');
+            const config = JSON.parse(runner.getOutput());
+            assert.ok(typeof config === 'object', "Expected config to be an object");
```
shalithasuranga, line 55: Could you please use `'` (single quotes) for strings? Thanks

spec/app.spec.js
```diff
@@ -79,6 +102,115 @@ describe('app.spec: app namespace tests', () => {
...
+                }
+            
+                function listener2() {
+                    listener2Triggered = true;
+                }
+            
+                await Neutralino.events.on('multiListenerEvent', listener1);
+                await Neutralino.events.on('multiListenerEvent', listener2);
+            
+                await Neutralino.app.broadcast('multiListenerEvent');
```
shalithasuranga, line 160: Are you sure this works always? Seems like `app.broadcast()` can end before the app trigger all the above listeners by keeping `false` for both variables. Do we need to add `__close()` within these listeners? Please research a bit more about this scenario

spec/app.spec.js
```diff
@@ -79,6 +102,115 @@ describe('app.spec: app namespace tests', () => {
...
+                }
+            
+                function safeListener() {
+                    safeListenerCalled = true;
+                }
+            
+                await Neutralino.events.on('errorEvent', errorListener);
+                await Neutralino.events.on('errorEvent', safeListener);
+            
+                await Neutralino.app.broadcast('errorEvent');
```
shalithasuranga, line 184: The same scenario as the above commented line. Please check

Sadaf-A:
> Thanks for sending this @Sadaf-A. This looks great. I've added some comments. We can merge this once we resolve them :tada:

Sure, I'll try to resolve them as soon as possible

spec/app.spec.js
Sadaf-A: Hey, since the listeners will be triggered in order I think adding `__close()` in the second listener should be a viable solution.

spec/app.spec.js
```diff
@@ -79,6 +102,114 @@ describe('app.spec: app namespace tests', () => {
...
+                }
+            
+                function listener2() {
+                    order.push(2);
+                }
+            
+                await Neutralino.events.on('orderEvent', listener1);
+                await Neutralino.events.on('orderEvent', listener2);
+            
+                await Neutralino.app.broadcast('orderEvent');
```
shalithasuranga, line 206: Thanks for updating the code according to notes. Btw, seems like here also we need to do the fix?

shalithasuranga (changes requested):
Now, there is only one comment. We can merge this PR as soon as that comment gets resolved. Thanks :tada:

Sadaf-A:
> Now, there is only one comment. We can merge this PR as soon as that comment gets resolved. Thanks 🎉

Hey, I have resolved the last comment 
Thank You

shalithasuranga (approved):
Thanks, looks great :tada:

PR #1262, opened 2024-06-11
GSoC2024: TestSuite - events API

Sadaf-A:
## Description

This PR adds the following testcases for the events API

## Changes proposed

 - on()
 - [x] triggers the event handler when an event is dispatched
 - [x] handles multiple events
 - [x] ensures event handlers are not duplicated
 - off()
 - [x] does not trigger the event handler after it is unset
 - [x] does nothing if trying to unset a handler that was not set
 - [x] can unset multiple handlers for the same event
 - dispatch()
 - [x] does not throw an error when no event listeners are registered
 - [x] triggers multiple callbacks for the same event
 - [x] does not trigger the callback after it is removed
 - [x] triggers the callback multiple times for multiple dispatches
 - [x] passes multiple data arguments to the callback
 - [x] handles errors in the callback gracefully
 - [x] continues to trigger remaining callbacks if one of them throws an error
 - broadcast()
 - [x] does not throw an error when no event listeners are registered
 - [x] triggers multiple callbacks for the same event
 - [x] does not trigger the callback after it is removed
 - [x] triggers the callback multiple times for multiple broadcasts
 - [x] passes multiple data arguments to the callback
 - [x] handles errors in the callback gracefully
 - [x] continues to trigger remaining callbacks if one throws an error
 
## How to test it

 - `cd spec && npm test events`

## Next steps

None.

## Deploy notes

None.

### DIFF
--- a/spec/events.spec.js
+++ b/spec/events.spec.js
@@ -6,18 +6,70 @@ describe('events.spec: events namespace tests', () => {
 
     describe('events.on', () => {
         it('sets an event without throwing errors', async () => {
-            let exitCode = runner.run(`
+            runner.run(`
                 function onTestEvent() {}
                 await Neutralino.events.on('testEvent', onTestEvent);
                 await __close('done');
             `);
             assert.equal(runner.getOutput(), 'done');
         });
+
+        it('triggers the event handler when an event is dispatched', async () => {
+            runner.run(`
+                let isEventTriggered = false;
+                function onTestEvent() {
+                    isEventTriggered = true;
+                    __close(JSON.stringify(isEventTriggered));
+                }
+                await Neutralino.events.on('testEvent', onTestEvent);
+                await Neutralino.events.dispatch('testEvent');
+            `);
+            const output = JSON.parse(runner.getOutput())
+            assert.strictEqual(output, true);
+        });
+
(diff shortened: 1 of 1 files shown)


spec/events.spec.js
```diff
@@ -6,30 +6,141 @@ describe('events.spec: events namespace tests', () => {
...
             `);
             assert.equal(runner.getOutput(), 'done');
         });
+
+        it('triggers the event handler when an event is dispatched', async () => {
+            runner.run(`
+                let isEventTriggered = false;
+                function onTestEvent() {
+                    isEventTriggered = true;
+                    __close(JSON.stringify(isEventTriggered));
```
shalithasuranga, line 22: Can we simply call `JSON.stringify({isEventTriggered: true})` without creating a new variable? Thanks :tada:

shalithasuranga (changes requested):
Thanks for sending this @Sadaf-A. I only have one comment regarding these changes. We can merge this after resolving it 🎉

spec/events.spec.js
Sadaf-A: Sure, I will make the changes

shalithasuranga (approved):
Everything looks great now. Thanks for sending your quick update :tada:

PR #1263, opened 2024-06-14
Disable the default context menu (right mouse click)

ratatoeskr666:
## Description
As discussed [here](https://github.com/neutralinojs/neutralinojs/discussions/630) I thought the ability to disable the default context menu would be useful. 
Unfortunately I can only test on windows so I would appreciate if someone could help me out with Linux/Mac.

## Changes proposed
Added an option to the neutralino config to enable/disable the context menu

## How to test it
- run Neutralino with this window option set: `enableContextMenu: false`
- no context menu opens when making a right mouse click

## Next steps
- Vote if this pull request should be merged as soon as the feature is completed
- Implement client library command
- Test on Linux/Mac

## Deploy notes
None.

### DIFF
--- a/api/window/window.cpp
+++ b/api/window/window.cpp
@@ -548,6 +548,7 @@ void __createWindow() {
                     windowProps.sizeOptions.maxWidth, windowProps.sizeOptions.maxHeight,
                     windowProps.sizeOptions.resizable);
     nativeWindow->setEventHandler(&window::handlers::windowStateChange);
+    nativeWindow->setEnableContextMenu(windowProps.enableContextMenu);
 
     #if defined(__linux__) || defined(__FreeBSD__)
     windowHandle = (GtkWidget*) nativeWindow->window();
@@ -848,7 +849,7 @@ json init(const json &input) {
         windowProps.enableInspector = input["enableInspector"].get<bool>();
 
     if(helpers::hasField(input, "borderless"))
-        windowProps.borderless = input["borderless"].get<bool>();
+        windowProps.borderless = input["borderless"].get<bool>();    
 
     if(helpers::hasField(input, "maximize"))
         windowProps.maximize = input["maximize"].get<bool>();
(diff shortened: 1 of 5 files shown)


ratatoeskr666:
Can somebody help me out?
For some reason the settings pointer is null in `webview.h:717` when I call the method via client routing:
```c++
void enableContextMenu(bool enable) override {
    ICoreWebView2Settings *settings = nullptr;
    HRESULT hr = m_webview->get_Settings(&settings);
    // Segmentation fault because of settings is nullptr!
    settings->put_AreDefaultContextMenusEnabled(enable);
  }
```

I'm not very common with C++ and I don't get why the webview is initialized and embedded but doesn't return a valid settings pointer. 

This is the client library code I added:
```js
export function enableContextMenu(enable: boolean): Promise<void> {
    return sendMessage('window.enableContextMenu', { enable });
};
```

ratatoeskr666:
Definitely but I think that's more a workaround than a clean solution.
The context menu from Neutralino is generated by the web view and should be disabled by the web view rather than just not passing the event to the webview.

Also starting from this PR my initial thought was to expand the functionality to provide custom context menus via Neutralinos API. Unfortunately I'm stuck with the whole C++ part and don't have much free time at the moment to dig deeper into this issue :/

thisjt:
I think you should reconsider implementing this. I'm not sure what the implications of fully disabling the context menu could have, as in the frontend, there are some implementations where the context menu event is being monitored and is overwritten by creating their own custom context menu. Some libraries could potentially lose this functionality.

What do you think? Besides, @honzamoser 's suggestion of disabling the context menu in the frontend is more than enough I think.

ratatoeskr666:
There is an API for WebView2 I wanted to expose. (https://learn.microsoft.com/en-us/microsoft-edge/webview2/how-to/context-menus)
Of course you can prevent right click event and (menu key press) in JS but since there is an API I thought it would be a good idea to use it instead and safe some end users some time.
Disabling the menu was just step 1. I originally wanted to customize my menu by removing some of the default items and ensuring default native behavior on the other hand. Maybe by using some sort of Neutralino Client Lib API like electron does (https://electronjs.org/docs/latest/api/menu#instance-methods). But I understand that this is obviously an edge case so I will close this PR since I'm stuck and I'm probably going to switch to electron anyway (for various reasons)

Issue #1073, opened 2023-02-20
Testing should not interrupt the machine.


NimishKashyap:
**Is your feature request related to a problem? Please describe.**
It is annoying to run all the tests in one go as it pops a new application window for each test. It interrupts other windows and tabs and the user basically have to sit and wait for the test to finish before doing anything practical.

**Describe the solution you'd like**
I want the test to run in the background instead of every time popping in-front of screen if whole test suite is set to run.

**Describe alternatives you've considered**
Of course the test suite can be run for each module at a time for minimizing the test time but it should not get interrupt the user's workflow. 

**Additional context**

shalithasuranga:
Closing this since this is not important at this moment of development. We can reopen this later if we need to. Thanks :tada:

PR #1270, opened 2024-06-28
[feature update] [Testing should not interrupt the machine] #1073

Manancode:
tests are set up to use runner.runInBackground for asynchronous execution, ensuring that the app.exit() function runs without blocking the main thread

runner.runInBackground is used to asynchronously fetch

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,23 @@
+{
+  "name": "neutralinojs",
+  "version": "1.0.0",
+  "description": "<div align=\"center\">   <picture>     <source media=\"(prefers-color-scheme: light)\" srcset=\"media/nllogo-light-theme.png\" width=\"300\"/>     <source media=\"(prefers-color-scheme: dark)\" srcset=\"media/nllogo-dark-theme.png\" width=\"300\"/>     <img alt=\"Neutralinojs logo\" src=\"media/nllogo-light-theme.png\" width=\"300\"/>   </picture>  </div>",
+  "main": "index.js",
+  "directories": {
+    "lib": "lib"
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -0,0 +1,23 @@
...
+    "type": "git",
+    "url": "git+https://github.com/Manancode/neutralinojs.git"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "bugs": {
+    "url": "https://github.com/Manancode/neutralinojs/issues"
+  },
+  "homepage": "https://github.com/Manancode/neutralinojs#readme"
```
Sadaf-A, line 22: Hey, I don't think a package.json file should exist in the root directory

spec/app.spec.js
```diff
@@ -5,7 +5,7 @@ describe('app.spec: app namespace tests', () => {
 
     describe('app.exit', () => {
         it('works without parameters', async () => {
-            let exitCode = runner.run(`
+            let exitCode = runner.runInBackground(`
```
Sadaf-A, line 8: Hey, `runner` does not expose any function called `runInBackground` If you have added that function please ensure that you have pushed that code because I don't see any changes made to `runner.js`

Sadaf-A (changes requested):
Hey, Thanks a lot for sending in this pull request. I have added a few comments please have a look

shalithasuranga:
Thanks for reviewing @Sadaf-A. Closing because the discussion is inactive. We can re-open this later if needed. Thanks :)

PR #1280, opened 2024-07-13
GSoC2024: TestSuite - storage API

Sadaf-A:
## Description
This PR adds test cases to the storage API for better coverage.

## Changes proposed

 - setData()
 - [x] throws an error for invalid datatypes
 - [x] persists data that is set
 - [x] updates existing data without throwing errors
 - [x] throws an error for empty string key
 - [x] sets large data values without throwing errors
 - [x] handles concurrent access without errors
 - getData()
 - [x] handles case-sensitive keys correctly
 - [x] handles concurrent access without errors
 - [x] gets JSON data without throwing errors
 -  getKeys()
 - [x] returns all stored keys

## How to test it

 - Run `cd spec && npm test storage`

### DIFF
--- a/spec/storage.spec.js
+++ b/spec/storage.spec.js
@@ -186,7 +186,7 @@ describe('storage.spec: storage namespace tests', () => {
             `);
             let keys = JSON.parse(runner.getOutput());
             assert.ok(Array.isArray(keys));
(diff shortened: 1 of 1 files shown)


shalithasuranga (approved):
Thanks so much for your contribution @Sadaf-A :tada:

PR #1288, opened 2024-07-26
Fix: Write Image to Clipboard on Windows

NayanPatidar:
## Description
 It fixes the working of writeImage() function In Windows 11.

## Before
Black Image is written to the clipboard.
Screenshot 2024-07-25 171312
## After
Image with all the specs is written properly to the clipboard
Screenshot 2024-07-25 171606
* The Latest image is the one written in the clipboard

## Changes proposed
 - Added the Alpha Mask and Alpha Shift to the image object in the Read Image Function
 - Set the Alpha Mask and Alpha Shift to the image spec in the Write Image Function

## How to test it
    async function WriteImage() {
     const image = await Neutralino.clipboard.readImage();
     console.log(image);
     const result = await Neutralino.clipboard.writeImage(image);
     console.log(result);
    }

### DIFF
--- a/api/clipboard/clipboard.cpp
+++ b/api/clipboard/clipboard.cpp
@@ -64,6 +64,8 @@ json readImage(const json &input) {
             { "redShift", spec.red_shift },
             { "greenShift", spec.green_shift },
             { "blueShift", spec.blue_shift },
+            { "alphaMask", spec.alpha_mask },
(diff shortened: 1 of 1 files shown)


NayanPatidar:
@shalithasuranga Any updates on this?

shalithasuranga (approved):
Thanks so much for sending this fix :tada:

PR #1304, opened 2024-08-17
GSoC2024: TestSuite - global variables

Sadaf-A:
## Description
This PR extends the test suite for global variables

## Changes proposed

 - [x] handles an unexpected NL_ARGS value
 - [x] handles NL_PID when not running in a process context
 - [x] handles NL_RESMODE with an unexpected value
 - [x] handles NL_EXTENABLED with various command-line arguments
 - [x] tests the mutability of NL_OS
 - [x] handles changing custom global variables
 - [x] verifies NL_PATH conforms to expected file path format
 - [x] should be able to access NL_OS in the global scope
 - [x] should be able to access NL_OS within a function
 - [x] should be able to access NL_OS within a nested function
 - [x] should not persist unintended modifications across tests

## How to test it

 - Run `cd spec && npm test globalvariables`

### DIFF
--- a/spec/globalvariables.spec.js
+++ b/spec/globalvariables.spec.js
@@ -1,5 +1,5 @@
 const assert = require('assert');
-
+const path = require('path');
 const runner = require('./runner');
 
 describe('globalvariables.spec: Global variables relatedtests', () => {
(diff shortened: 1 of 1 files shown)


shalithasuranga (approved):
Thanks so much @Sadaf-A for sending this pull request :tada:

PR #1306, opened 2024-08-27
Removed _gsoc-win7-support directory

BhuvanAde:
moved the `_gsoc-win7-support` dir to a diff repo;


shalithasuranga (approved):
Thanks @BhuvanAde :tada:

PR #1317, opened 2024-09-22
Fix #1316

### LINKED ISSUE #1316: ID Collision in `os.spawnProcess` method
Using `spawnedProcesses.size()` to determine `virtualPid` can lead to multiple processes receiving the same Neutralino-scoped ID. This can cause 'spawnedProcess' events to trigger handlers across multiple processes unintentionally.

**How it can happen:**
1. Three processes are spawned: their IDs are `[0, 1, 2]`
2. Process IDs `[0, 2]` are killed, leaving only `[1]`
3. A new process is spawned while the size of the map is 1, resulting in `[1, 1]`
4. The 'exit' event is called with ID 1. Panik!

**The offending bit of code:**
https://github.com/neutralinojs/neutralinojs/blob/2655575247b301d6b4a1fc68c8a0e1d9a1bbb1a6/api/os/os.cpp#L133-L164

**Related:**
https://github.com/neutralinojs/neutralino.js/issues/64

benjammin4dayz:
## Description

This change aims to eliminate ID collision in the `os.spawnProcess()` method. See #1316

## Changes proposed

- Introduce an incrementing counter in lieu of `map::size()` for virtual process IDs
- Add a test case to check for unique IDs

## How to test it

- Run the new test case added to `os.spec.js`

## Next steps

None.

## Deploy notes

None.

### DIFF
--- a/api/os/os.cpp
+++ b/api/os/os.cpp
@@ -12,6 +12,8 @@
 #include <cstring>
 #include <vector>
 #include <filesystem>
+#include <atomic>
+#include <climits>
 
 #include "resources.h"
 #include "lib/tinyprocess/process.hpp"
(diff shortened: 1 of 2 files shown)


shalithasuranga (approved):
Thanks so much for your contribution @benjammin4dayz :tada:

benjammin4dayz:
My pleasure! Thanks for reviewing.

PR #1332, opened 2024-11-05
Fix neu devtools extension condition

benjammin4dayz:
## Description

This change will allow the devtools extension to connect when the Neutralino app is in development configuration.

## Changes proposed

- Flip the conditional statement in `settings.cpp` to `!resources::isBundleMode()`

## How to test it

 - Run a frontend library app that requires a devtools connection to patch the HTML file

## Next steps

None.

## Deploy notes

None.


shalithasuranga (approved):
Thanks so much for this fix @benjammin4dayz. This patch fixes the broken auto-reload feature in `neu run` :tada:

PR #1345, opened 2024-12-20
fixed bug causing syntax error while running build cmd

Bxpx6:
The issue  (SyntaxWarning: invalid escape sequence '\ ') is fixed.  By making the string literal raw by adding an r prefix to the triple-quoted string. Hope that solves the bug.

### DIFF
--- a/bin/resources/js/main_spec.js
+++ b/bin/resources/js/main_spec.js
@@ -1,0 +1,37 @@
+
+
+
+Neutralino.init();
+
+Neutralino.events.on("ready", async () => {
+    await __init();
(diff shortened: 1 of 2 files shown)


shalithasuranga (changes requested):
Hello @Bxpx6, thanks so much for sending this pull request. Could you please remove `main_spec.js` changes and commit only `bz.py`? Then, we can merge this PR.

Thanks :tada:

Bxpx6:
Done. Removed main_spec.js and commit only  bz.py done.

PR #1346, opened 2024-12-22
commiting bz.py and fixing the bug that caused syntax error/Bxpx6

Bxpx6:
## Changes proposed

The bug is fixed 
The issue (SyntaxWarning: invalid escape sequence '\ ') is fixed. By making the string literal raw by adding an r prefix to the triple-quoted string. Hope that solves the bug.


shalithasuranga (approved):
Thanks so much for sending this fix @Bxpx6

Issue #1291, opened 2024-07-29
Configuration flag to allow using `file://` with local files


CosmoMyzrailGorynych:
**Is your feature request related to a problem? Please describe.**
Right now, if you link to any file with the `file://` protocol, it will log an error `Not allowed to load local resource: file:///yadayada/img/splash.png`. This complicates developing applications that work with user files and need to display or process local images, sounds, videos, etc.

**Describe the solution you'd like**
A flag in `neutralino.config.json` that allows use of local files.

**Describe alternatives you've considered**
I can read the file with `Neutralino.filesystem.readBinaryFile`, make a blob from it, and get a URL for it, but that's not developer-friendly.

**Additional context**
For Chrome-like browsers, this is controlled with the `--allow-file-access-from-files` command line flag, which should solve the issue for Windows + Edge. I don't know how things work on Linux and Mac, though.

Patrikgrinsvall:
This i would gladly pay some sponsorship for! Now the process of loading local files is somewhere between quite complex and "ugly-hackish". It often involves custom events or extensions and transferring data as some serialized or ASCII representation of binary data, converting it back and forth, just to see it fail intermittently due to bugs in underlying libraries. Nightmare to debug in an app released to end-users.

As a matter of fact, its so cumbersome i usually copy the files into the publicly available docroot dir and then read from there in front-end. Also i never remove them or cleanup fails. But, this approach will never save any rain forest since duplicated files on disk is slowly becoming a monster of an evil resource thief and just the other day i was wondering if my 1.4tb disk went "AWOL" instead of swapping tmpfs . Almost like running your car on gas filled with tar and asphalt just for the smell... or maybe not but anyway a very missing feature.

Note! If i recall this is not unique to neutralino but is also an annoying task in other frameworks and often requires plugins or extensions to get it decent but at the cost of complexity and bugs and more gray hair than most of us want.

CosmoMyzrailGorynych:
> It often involves custom events or extensions and transferring data as some serialized or ASCII representation of binary data, converting it back and forth, just to see it fail intermittently due to bugs in underlying libraries

Idk what hellish abominations you made, but loading local files in Neutralino right now is doable with this, without any extensions:

```ts
const fs = Neutralino.filesystem;
const getLocalFileLink = async (file: string): Promise<string> => {
  const arrayBuffer = await fs.readBinaryFile(file);
  const blob = new Blob([arrayBuffer]);
  return URL.createObjectURL(blob);
[...]

PR #1356, opened 2025-01-04
✨ app.mount, app.unmount for hosting files from filesystem

### LINKED ISSUE #993: Serving local files, bypassing file:// limitation
Some application require direct access to local user files directly from web view. Example use cases: image viewers, media players.

Ofcourse we can use the filesystem module to load the data to a blob and make an URL out of it, or copy it to local resources (seems like a bad idea).
These solution make sense only for smaller file and when you need just a couple of files. If on the other hand you need dozens of files, when for example creating lots of small thumbnails, things get out of hand very fast.

Similar issue was already reported in #468.

Since removing security related to preventing loading from file:/// is not smart, I propose serving those files from the built in http server, similarly like the assets are being served.

Probably types of files being server should be limited, only to those that can actually be directly displayed within a webview (media types only?).

My idea would be to create a new module `Neutralino.serve.' with some functions:
- `.enable(true/false)` - enable or disable serving local files
- `.prefix('__local')` - provide an URL prefix for serving local files

The idea would be that files under '/{prefix}/' are served from root of the file system OR use user provided mapping, requiring programmer to first define each URL mapped to local file, eg.

```
Neutralino.serve.add('/__local/my-image.jpeg', '/Users/XYZ/Documents/my-image.jpeg');
```

I was considering adding an extension, but this feature seems kind of standard, and creating a separate binary just to handle this kind of situation seems off.

**Additional context**
N/A

I am willing to try implementing this if such feature could be considered for merging into master.

CosmoMyzrailGorynych:
Closes #993
Provides a solution to #1291 use case but is not a solution for the `file://` protocol.

## Description
This PR adds new methods `Neutralino.app.mount` and `Neutralino.app.unmount` that allow to add and remove filesystem folders on local machine to be included in Neutralino's web server routing.

Calling `Neutralino.app.mount('/sausage', 'C:/Users/CoMiGo/Some Directory');` will make any requests to `/sausage/**` to read and serve files from that folder:

* fetching `/sausage` will try to load `C:/Users/CoMiGo/Some Directory/index.html`
* fetching `/sausage/hello.txt` will try to load `C:/Users/CoMiGo/Some Directory/hello.txt`
* fetching `/not-a-sausage/hello.txt` will try to load a file from the app's resources folder as usual.

Calling `Neutralino.app.unmount('/sausage')` will revert the changes to the router.

Several different mounts can be established at once, too.

This unlocks several use cases for app developers:

* Scoping a part of end user's filesystem as a subroute in an app to:
  * Simplify working with filesystem's paths (which is useful for anything that works with user projects);
  * Bypass issues with `file://` protocol that is currently unusable and would have problems with WebAudio and other issues anyways;
  * Remove the need to use filesystem.readBinaryFile + Blobs to use local images/audio/video in webpages;
* Local testing for web projects built by an app: for example, games and websites in a game maker / website constructor. Can be made with an iframe or by calling `app.mount`, then navigating to it with `location.href = '/sausage';`.
* Proxying the whole filesystem for image viewers and other file-manager-like apps. 

## Changes proposed

* Add app.mount and app.unmount methods
* Add new error codes relevant to the methods
* Add tests for them
* Remove trailing whitespace in a couple of places

## How to test it

I added a couple tests to `app.spec.js`. Should be doable with `npm run test` in the spec folder

## Next steps

* Send a PR to the client library (done, https://github.com/neutralinojs/neutralino.js/pull/142)
* Send a PR for the docs (done, https://github.com/neutralinojs/neutralinojs.github.io/pull/367)

## Deploy notes

⚠️ This is my first real C++ work **ever**, ignoring the long forgotten university curriculum. Please check the code carefully as I may or may not have done dumb shit. And if not Pieces App and Claude Sonnet, I wouldn't write it in one evening. _But,_ it works, I tested it with different paths and made sure they are normalized.

You are also free to move stuff around and/or rename the methods.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 ![Build status](https://github.com/neutralinojs/neutralinojs/actions/workflows/test_suite.yml/badge.svg)
 [![FOSSA Status](https://app.fossa.com/api/projects/git%2Bgithub.com%2Fneutralinojs%2Fneutralinojs.svg?type=shield)](https://app.fossa.com/projects/git%2Bgithub.com%2Fneutralinojs%2Fneutralinojs?ref=badge_shield)
 
-Neutralinojs is a lightweight and portable desktop application development framework. It lets you develop lightweight cross-platform desktop applications using JavaScript, HTML and CSS. Apps built with Neutralinojs can run on Linux, macOS, Windows, Web, and Chrome. Also, you can extend Neutralinojs with any programming language (via extensions IPC) and use Neutralinojs as a part of any source file (via child processes IPC). 
+Neutralinojs is a lightweight and portable desktop application development framework. It lets you develop lightweight cross-platform desktop applications using JavaScript, HTML and CSS. Apps built with Neutralinojs can run on Linux, macOS, Windows, Web, and Chrome. Also, you can extend Neutralinojs with any programming language (via extensions IPC) and use Neutralinojs as a part of any source file (via child processes IPC).
 
 - [Neutralinojs vs. Electron, NW.JS, Tauri, NodeGui, Flutter, .Net MAUI, Wails](https://github.com/Elanis/web-to-desktop-framework-comparison)
 - [Neutralinojs vs. Electron vs. NW.js (2018)](https://github.com/neutralinojs/evaluation)
@@ -21,14 +21,14 @@ Neutralinojs is a lightweight and portable desktop application development frame
 Get started with the neu CLI:
 
 ```bash
- # Creating a new app
- npm i -g @neutralinojs/neu
- neu create hello-world
- cd hello-world
- neu run
- 
- # Building your app (No compilation - takes less than a second)
- neu build
+# Creating a new app
+npm i -g @neutralinojs/neu
+neu create hello-world
+cd hello-world
+neu run
+
+# Building your app (No compilation - takes less than a second)
+neu build
 ```
 
 You can use your favorite frontend frameworks:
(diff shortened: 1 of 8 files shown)


shalithasuranga (approved):
Hello @CosmoMyzrailGorynych This will be a great feature as you described :tada: Thanks so much for sending this complete PR. I think only a few adjustments need to be done in the code. Also, I think it's better to add mappings to the config file -- not as functions since app developers may need mappings before even calling these mounting functions. 

I'll let you know after making my changes (loading from config, etc.) for your verification :tada:

shalithasuranga:
Hello @CosmoMyzrailGorynych, seems like it's better to add these functions first and add the config file support later since the config approach limits using dynamic mappings. We can introduce config file support later :tada:

CosmoMyzrailGorynych:
> Also, I think it's better to add mappings to the config file -- not as functions since app developers may need mappings before even calling these mounting functions.

Dynamic routes are a must — in my app, for example, I don't know beforehand which directories to mount, as a user can open a project from anywhere. Thus the new API instead of configuration options like, say, socketsupply has. Static routes may be a useful addition, but I don't foresee any use cases where API calls would not do the same thing. If you know you need to serve specific mounts with specific files, why not bundle it into app's `resources.neu`?

For routes in configuration files, what can be useful is support for flags like `$HOME`, `$CWD`, and maybe other paths from [os.getPath](https://neutralino.js.org/docs/api/os#osgetpathname) API. Still wouldn't suit my own needs but maybe will be helpful for others.

CosmoMyzrailGorynych:
> I hope your implementation approach stays the same in the modified code ;)

Yup, everything is fine. Thank you for the refactoring.

PR #1358, opened 2025-01-07
:zap: Host pages from 127.0.0.1 instead of localhost

### LINKED ISSUE #957: Windows builds are suspiciously slow
**Describe the bug**
I am working on a crossplatform open source app. An offline app with some simple pages.
On macOS and Linux, the loading process is super fast. 
But on Windows 10 and 11 it takes 4-5 seconds for a page to open.

**To Reproduce**
1. Create an empty project.
2. Create a production build.
3. Open the app on Windows 10/11.
4. Open the inspector.
5. Open the network tab.
6. A single file request has (in my case) 20 times longer loading time than on MacOS/Linux.

**Expected behavior**
A request should only take a few milliseconds for an offline app, not half a second.
Sure, I can tweak the app to not have as many file requests. But that shouldn't be necessary if everything is offline.

macOS:
macOS
Windows:
Windows

**Specifications**  
 - OS: Windows 11 x64, Windows 10 x64
 - Neutralinojs version: v4.6.0
 - Neutralinojs client library version: v3.5.0
 - Neutralinojs CLI version: v9.3.1

**Additional context**
- WebView2-Version: 103.0.1264.62 (Official build) (64-bit)

### LINKED ISSUE #1217: Faster startup
I don't have exact metrics, but it seems it takes around 1-1.5 seconds to go from launch to render a basic hello-world app with just a static html file. Nothing complex yet – and it still around 1.5 seconds just to load a single static html...

There's also the white background problem, but the background is not really the issue as I described a workaround here:

- https://github.com/neutralinojs/neutralinojs/issues/814#issuecomment-1955426351
  > One workaround is to put in config: `"hidden": true`, then in your application run `Neutralino.window.show()` asap

I don't know the code to fix the startup situation, but I kinda hope that it is possible to run static files faster. It's unreal to spend 1.5 sec just to load statics. For just `.html` statics, I would expect no more than 0.2 sec to start the app.

Electron???? 😒
Maybe even non-js... Rust? 😮
Of course, Rust isn't really an option... I would spend ages to get used to Rust language. It runs fast, though...

**Additional context**
I'm sorry for not having exact metrics at hand, but VSCode (written in Electron) seems to show some initial UI in just 0.2-0.4 sec, and it loads entire code editor with bunch of files in total of 1.5 sec... That's an entire complex app in 1.5 sec! While in neutralino-app, we spend 1.5 sec just for one single static `.html` hello-world kind of an app 😒 So I'm pretty sure, as fatty as Electron is, it can start up from a static html file – much faster than Neutralino. For a tiny app of 2 megabytes, I would so much expect it to startup instantaneously... But currently, Neutralino isn't capable to do this: no matter how tiny it seems, it requires hell lot of a time to startup.

CosmoMyzrailGorynych:
Closes #957
Closes #1217

## Description
Ssssoooo... Windows is a "miracle".
When serving files from `localhost`, any [requests to `localhost` will go through DNS resolve system](https://stackoverflow.com/questions/2386299/running-sites-on-localhost-is-extremely-slow) and will have long Initial Connection time. In my case this time takes 300ms for **each** request on a pretty powerful laptop.

The problem became evident when migrating to my Neutralino.app.mount method when working with local files.

Loading a project on localhost:

localhost

See those yellow lines? Each is 300ms+ long and is resolving `localhost` (the Initial Connection stat in DevTools' Network tab.)

Loading the same project on 127.0.0.1:

Changing the hostname also made my app load noticeably faster.

## Changes proposed

* Neutralino.js will now open applications on `http://127.0.0.1:PORT` instead of `http://localhost:PORT`

## How to test it

* Apps work even despite having a client patched to connect through websockets with 127.0.0.1. Speed change should be measured manually.

## Next steps

* (Done) Add a PR to neutralino.js client library as it currently opens its websockets connection on `localhost`. This will make native API initialization faster on Windows. https://github.com/neutralinojs/neutralino.js/pull/143

## Deploy notes

⚠️ This may break some existing projects if they defined a [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) in their apps. (For example, to load and work with blobs.)

In my engine, I had to expand it a bit:

```diff
<meta
    http-equiv="Content-Security-Policy"
    content="
        connect-src
-            https: file: ipc: socket: ws://localhost:* http://localhost:* blob:;
+            https: file: ipc: socket: ws://localhost:* ws://127.0.0.1:* http://localhost:* http://127.0.0.1:* blob:;
        script-src
-            https: socket: http://localhost:* 'unsafe-eval' 'unsafe-inline' blob:;
+            https: socket: http://localhost:* http://127.0.0.1:* 'unsafe-eval' 'unsafe-inline' blob:;
        img-src
-            https: data: file: blob: http://localhost:*">
+            https: data: file: blob: http://localhost:* http://127.0.0.1:*;">
```

But default Neutralino project runs just fine.

### DIFF
--- a/server/neuserver.cpp
+++ b/server/neuserver.cpp
@@ -133,7 +133,7 @@ string init() {
         settings::setPort(port);
     }
 
-    string navigationUrl = "http://localhost:" + std::to_string(port);
+    string navigationUrl = "http://127.0.0.1:" + std::to_string(port);
     json jUrl = settings::getOptionForCurrentMode("url");
 
     if(!jUrl.is_null()) {


shalithasuranga (approved):
Hello @CosmoMyzrailGorynych, thanks so much for sending this pull request with a detailed explanation and screen recordings :tada:

PR #1365, opened 2025-01-22
add setVDocRoot function

stxh:
## Description
Implements a virtual document root for local folders. When building an application to access local files, system functions can be used to read and display file content. However, if a file (such as an HTML document) references other local files (e.g., images in the same folder), these references typically fail to resolve (see #1291).

## Changes proposed
Adds a `setVDocRoot` function to designate a local folder as the virtual document root for resources.

## How to test it
Before opening a local HTML file containing referenced images, use `setVDocRoot` to set the virtual document root to the folder containing the HTML file and images. Verify that the images are displayed correctly.

## Deploy notes
This change introduces a new function, `setVDocRoot`, which may require adjustments to existing build processes or deployment scripts if they rely on specific file paths. Ensure that the correct path is passed to setVDocRoot during application initialization. Consider adding documentation or examples demonstrating its usage.

You need update your neutralino.js too. https://github.com/neutralinojs/neutralino.js/pull/147

## Usage
```js
      Neutralino.server
         .setVDocRoot(path)
         .then((resault) => {
            console.log("setVDocRoot:", resault)
        })
        .catch((error) => {
          console.error("setVDocRoot error:", error);
        });
```

### DIFF
--- a/api/server/server.cpp
+++ b/api/server/server.cpp
@@ -55,5 +55,23 @@ json getMounts(const json &input) {
     return output;
 }
 
+//  set a Virtual Document Root
+json setVDocRoot(const json &input) {
+    json output;
+    string path = input["path"].get<string>();
+    errors::StatusCode mountStatus = router::setVDocRoot(path);
+    
+    if(mountStatus != errors::NE_ST_OK) {
+        output["error"] = errors::makeErrorPayload(mountStatus, path);
+    }
+    else {
+        output["message"] = "vdocroot mounts to " + path ;
+        output["success"] = true;
+    }
+
+    return output;
+}
+
+
 } // namespace controllers
 } // namespace server
--- a/api/server/server.h
+++ b/api/server/server.h
@@ -15,6 +15,9 @@ json mount(const json &input);
 json unmount(const json &input);
 json getMounts(const json &input);
 
+// set a Virtual Document Root
+json setVDocRoot(const json &input);
+
 
 } // namespace controllers
 
--- a/server/router.cpp
+++ b/server/router.cpp
@@ -146,6 +146,7 @@ map<string, router::NativeMethod> methodMap = {
     {"server.mount", server::controllers::mount},
     {"server.unmount", server::controllers::unmount},
     {"server.getMounts", server::controllers::getMounts},
+    {"server.setVDocRoot", server::controllers::setVDocRoot},
     // Neutralino.custom
     {"custom.getMethods", custom::controllers::getMethods},
     // {"custom.add", custom::controllers::add} // Sample custom method
@@ -261,6 +262,23 @@ map<string, string> getMounts() {
     return mountedPaths;
 }
 
+// add vdocroot
+string vdocroot = string();
+
+errors::StatusCode setVDocRoot(string &target) {
+    target = helpers::normalizePath(target);
+
+    if(!filesystem::exists(target)) {
+        return errors::NE_FS_NOPATHE;
+    }
+    if(!filesystem::is_directory(target)) {
+        return errors::NE_FS_NOTADIR;
+    }
+    
+    vdocroot = target;
+    return errors::NE_ST_OK;
+}
+
 router::Response getAsset(string path, const string &prependData) {
     router::Response response;
     vector<string> split = helpers::split(path, '.');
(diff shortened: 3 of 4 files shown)


CosmoMyzrailGorynych:
What benefit does this bring compared to server.mount command? This doesn't solve the issue with the file:// protocol and I think server.mount('/', fsPath) will do the same as your PR. If it is another workaround then it's not needed, and file:// protol is needed due to memory concerns, to prevent memory pressure due to caching http resources.

stxh:
> What benefit does this bring compared to server.mount command? This doesn't solve the issue with the file:// protocol and I think server.mount('/', fsPath) will do the same as your PR. If it is another workaround then it's not needed, and file:// protol is needed due to memory concerns, to prevent memory pressure due to caching http resources.

After use server.mount('/'. fsPath), you can't reload other js or other resource files from resources.neu

CosmoMyzrailGorynych (changes requested):
After further looking, my two cents are: this is unneeded bloat and is likely insecure. 

1. Bloat: `server.mount` does the same thing in a more controlled, safer way, and as a developer you don't really care whether you use filesystem files from `/` or `/whatever/`. You're the boss and define what loads from where. I'm against duplicated APIs as it means higher load for maintenance, and also will cause confusion among users.
2. Insecure: in case your app links to a file that does not exist in resources.neu (e.g. due to a bug or exploit like changing user settings to load an invalid translation or theme), it will try to load an arbitrary file. Damage and severity will depend on each app and developer, but let's be mindful and not allow such cases at all.

Also:

* This PR does not solve the issue with file:// protocol (#1291). It proposes another solution to an already closed issue #993. 1291 is a distinct issue due to being an obvious browser feature with specific cache handling.
* I see no benefit in using `/` as a base route for filesystem requests aside from writing a few less characters in your code.
* Method name is not quite intuitive and Neutralino doesn't usually use abbreviations and shortenings in them.

shalithasuranga:
Hello @stxh, thanks so much for sending this pull request. Seems like this feature tries to offer what the `Neutralino.server` namespace offers. The `Neutralino.server` namespace has not yet been released. Could you please read the changelog.md file and check whether the `Neutralino.server` solves your requirement. If it does, we can close this PR. Otherwise, we can improve the existing `server.*` functions to satisfy your needs.

Thanks so much for reviewing this PR @CosmoMyzrailGorynych :tada:

stxh:
> 1. Bloat: `server.mount` does the same thing in a more controlled, safer way, and as a developer you don't really care whether you use filesystem files from `/` or `/whatever/`. You're the boss and define what loads from where. I'm against duplicated APIs as it means higher load for maintenance, and also will cause confusion among users.

I tested `server.mount`, but it caused the application's display to malfunction.

> 2. Insecure: in case your app links to a file that does not exist in resources.neu (e.g. due to a bug or exploit like changing user settings to load an invalid translation or theme), it will try to load an arbitrary file. Damage and severity will depend on each app and developer, but let's be mindful and not allow such cases at all.

`resources.neu` is an ASAR archive. Its contents can be easily extracted and repackaged, offering no inherent security.

CosmoMyzrailGorynych:
> I tested server.mount, but it caused the application's display to malfunction.

Elaborate. I build a huge app with it and it works just fine.

> resources.neu is an ASAR archive. Its contents can be easily extracted and repackaged, offering no inherent security.

This is not an excuse to worsen the situation.

CosmoMyzrailGorynych:
@stxh Look, let's rehearse what's happenning:

* You're introducing an API that duplicates existing functionality. You might have overlooked it as separate developers work separately, but this is not a reason to push it further — every method needs to be maintained and kept in mind while developing new functions and thus every new method is additional workload for an indefinite amount of time.
  * You still haven't explained why one would **require** hosting user's filesystem at the `/` route and why filesystem and Neutralino's routes must be merged. With recently introduced `server` namespace, developers in Neutralino have the flexibility to build an app with custom routes pointing from any filesystem folder to any subroute, and as a developer of a huge app on Neutralino I can't imagine the use cases where there are limitations with that.
* Your solution is insecure. Neutralino.js may introduce code signing and/or code protection in the future, while your method will persist and create a security hole. This is not right.
* Your PR will close an issue created by me while not providing a solution to it. You're not doing anything with the `file://` protocol. Obviously I'm against that as it essentially mutes the issue without a proper solution.
* Your PR also got messier and now has changes to create_release workflow, pulls in your own workflow, magically introduces Bun in readme, etc. Though I admire love for Bun, this PR won't be merged. It is now a mess.
* You also haven't addressed code style issues (method naming).

I don't think your PR is needed. I think you should close it. Please use your time wisely.

shalithasuranga:
I'll close this pull request since the same feature is available in the `server` namespace. Thanks so much for sending this pull request and starting a discussion :tada:

Issue #1367, opened 2025-01-25
Improve generateToken() Function to Fix Token Repetition and Enhance Randomness


SharonIV0x86:
**Bug**
The ``generateToken()`` function does not works as expected, as it generated the same token again and again due to the fact that the ``srand(time(NULL))`` function is being called every time ``generateToken()`` is invoked. This means that the random seed is being initialized with the same value ``(time(NULL))`` because ``time(NULL)`` doesn't change within the same second.

Additionally the current implementation of ``generateToken()`` function is inefficient for random generation, this can be replaced with a better and modern implementation for random token generation which utilizes ``std::mt19937`` and ``uniform_int_distribution<>``.  

I'd like to work on improving the ``generateToken()`` function, kindly assign this issue to me. 

**Steps To Reproduce**

I have taken the exact implementation of the ``generateToken()`` function from [here](https://github.com/neutralinojs/neutralinojs/blob/2c386620486123639aee3aa69a7ed9e620aedc95/helpers.cpp#L38) and i have attached the link to the online compiler to demonstrate [this](https://onecompiler.com/cpp/4373zjzr8) issue. In the demonstration you can see that even with subsequent calls to the ``generateToken()`` function the same token is generated each time, the randomness changes after 1 second. 

As far as i know there isn't a exact or straightforward way to reproduce this problem. 

But you can pass this flag ``--export-auth-info=true`` to the neu CLI ``neu run`` or manually calling the ``./bin/neutralino-linux_x64`` to view the auth token

 
**Expected behavior**
Each time the ``generateToken()`` function is called it should generate a new token completely different than the previous one. 

**Specifications**  
 - OS: [Manjaro Linux]
 - Neutralinojs version: [v5.5.0]
 - Neutralinojs client library version: [v5.5.0]
 - Neutralinojs CLI version: [v11.3.0]

SharonIV0x86:
I'd like to work on this issue, kindly assign it to me.

SharonIV0x86:
@shalithasuranga can i work on this?

shalithasuranga:
Thanks @SharonIV0x86 for posting this. Sure, looking forward to see your pull request for this 🎉  happy coding

PR #1371, opened 2025-01-28
Fixed token generation in generateToken() function in helpers.cpp

SharonIV0x86:
## Description
This PR fixes the ``generateToken()`` method in ``helpers.cpp`` the generated token is used for authentication. The current implementation of this function generates the same token again and again as mentioned in this issue https://github.com/neutralinojs/neutralinojs/issues/1367 The issue was the way ``srand(time(NULL))`` was placed or being used, but this method for random generation is old and outdated and should not be used.

Fixes: #1367 

## Changes proposed
- Get rid of ``srand(time(NULL))``
- Used standard provided [Mersenne Twister](https://en.wikipedia.org/wiki/Mersenne_Twister) for random generation

## How to test it
The testing is not as straight forward, there are no tests available in ``/spec`` folder to test the efficiency of ``generateToken`` function, but you can view the comparison by running the old implementation of the ``generateToken()`` function versus the new implementation proposed in this PR from the online compiler link mentioned below.

### Old / Before
[here](https://onecompiler.com/cpp/4373zjzr8)
As you can see in the output the tokens generate after each subsequent calls are same and collisions rate is **99.99%**
### New / After
[here](https://onecompiler.com/cpp/437dtktbg)
As you can see in the output the tokens are all random and the collisions rate is **0.001%** indicating very good performance.

### DIFF
--- a/helpers.cpp
+++ b/helpers.cpp
@@ -50,9 +50,12 @@ string generateToken() {
     std::uniform_int_distribution<size_t> distribution(0, characters.size() - 1);
 
     for (size_t i = 0; i < length; ++i) {
+        if (i == 47) {
+            token += ".";
+        }
         token += characters[distribution(generator)];
     }
-
+    
     return token;
 }
 
--- a/helpers.cpp
+++ b/helpers.cpp
@@ -37,19 +37,22 @@ vector<string> splitTwo(const string &s, char delim) {
 }
 
 string generateToken() {
-    static const std::string characters =
-        "-_0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
+    static const string characters =
+        "-_"
+        "0123456789"
+        "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+        "abcdefghijklmnopqrstuvwxyz";
 
     const int length = 96;
 
-    std::string token;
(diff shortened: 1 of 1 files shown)


shalithasuranga (changes requested):
Hello @SharonIV0x86, thanks so much for sending this pull request. Could you please verify this implementation's output with the previous implementation's output? `NL_TOKEN` has two token strings separated with a dot (`.`) character. It seems like the new implementation doesn't include it in its algorithm.

We can merge this once you update the new algorithm and provide some sample output tokens.

Thanks :tada:

SharonIV0x86:
> Hello @SharonIV0x86, thanks so much for sending this pull request. Could you please verify this implementation's output with the previous implementation's output? `NL_TOKEN` has two token strings separated with a dot (`.`) character. It seems like the new implementation doesn't include it in its algorithm.
> 
> We can merge this once you update the new algorithm and provide some sample output tokens.
> 
> Thanks 🎉

I must have overlooked that, i can incorporate this in the new implementation as well, it seems like the older implementation has a ``.`` (dot) character after 47 characters in the token, I'll add this in the new implementation and provide a sample output. Thank you for reviewing.

shalithasuranga (approved):
Thanks so much for doing these updates. Looks great now :tada:

Issue #1373, opened 2025-01-29
Allow to specify environment variables for `os.spawnProcess`


ddorn:
**Is your feature request related to a problem? Please describe.**
I need to run processes that expect some environment variables to be set.

**Describe the solution you'd like**
Either another function such as `os.spawnProcessWithEnv(command, env, cwd)` where `env` is any object such as `{"VAR": "value"}`,
or adding a last parameter to `os.spawnProcess`.

**Describe alternatives you've considered**
It's already possible to do so by running commands such as `env VAR=value sh -c 'escaped command'`, but can be more tricky, because of the escaping and the need to be cross-platform.

**Additional context**
TinyProcessLib has support for this feature, as can be seen in https://gitlab.com/eidheim/tiny-process-library/-/blob/master/examples.cpp#L151

Thank you for creating and maintaining this amazing library!

SharonIV0x86:
@ddorn So when you call ``os.spawnProcessWithEnv(command, env, cwd)`` with env as
```json
{
    "VAR": "value"
}
```
you'd like the VAR to be created in the environment variable if it does not exist? and if it exists then the process should continue normally?

SharonIV0x86:
**@ddorn @shalithasuranga** I'd love to work on this issue, but I have a few questions before proceeding. One key point is handling cases where the specified environment variable does not exist—should we create it with the provided value?  


To simplify things, we could introduce a new API method, such as **`Neutralino.os.createEnvVar(name, value)`**, to explicitly allow setting environment variables. However, I’d like to understand the security implications of this feature. While it could be misused to create and modify environment variables excessively, similar behavior is already possible through **`os.execCommand(command, options)`**.  But i dont think this should by any concerning.


That said, adding **`createEnvVar`** would give developers more control over environment variables in a straightforward way. Additionally, this function could serve as the foundation for implementing **`os.spawnProcessWithEnv(command, env, cwd)`**, making it easier to run processes with custom environment variables.

ddorn:
Thank you for looking at this!

What you are suggesting is not quite what I had in mind. If I understood correctly you're suggesting to change the environment of the main process before spawning a new one? Instead, what is usually done is to keep the current environment unchanged, but spawn the process with its own, modified environment. See for example `execle` and `execvpe`.

SharonIV0x86:
> Thank you for looking at this!
> 
[...]

PR #1374, opened 2025-01-30
Allow to specify environment variables for os.spawnProcess

SharonIV0x86:
Addressing https://github.com/neutralinojs/neutralinojs/issues/1373

## Description
This PR adds a function ``spawnProcessWithEnv`` to spawn a process with a custom set of environment variables, the process will only be able to use those specified environment variables only.

## Changes proposed
- ``spawnProcessWithEnv`` to spawn a process with a list of custom environment variables to use for that process  
- Added proper error handling.
- Registered functions in the ``routers.cpp``

## Next steps
- Take input from the maintainer
- Improve the function
- Make further required changes.

### DIFF
--- a/api/os/os.cpp
+++ b/api/os/os.cpp
@@ -710,27 +710,67 @@ json getPath(const json &input) {
     }
     return output;
 }
-string setEnvVar(const std::string& key, const std::string& value) {
-    json output;
-    if(!getEnv(value).empty()){
-        output["error"] = errors::makeErrorPayload(errors::NE_OS_ENVEXISTS);
-    }else if(key.empty() || value.empty()){
-        output["error"] = errors::makeMissingArgErrorPayload();
+
+pair<int, int> spawnProcessWithEnv(const string& command, const map<string, string>& env, const string& cwd) {
+    unordered_map<string, string> env_unordered(env.begin(), env.end());
+
+    TinyProcessLib::Process* childProcess;
+    lock_guard<mutex> guard(spawnedProcessesLock);
+
+    int virtualPid = nextVirtualPid++;
+    if (nextVirtualPid == INT_MAX) {
+        nextVirtualPid = 1;  
     }
 
-#ifdef _WIN32
-    if(_putenv_s(key.c_str(), value.c_str()) != 0){
-        output["error"] = errors::makeErrorPayload(errors::NE_OS_CNCENV);
+    childProcess = new TinyProcessLib::Process(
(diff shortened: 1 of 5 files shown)


SharonIV0x86:
I have implemented a new function called ``spawnProcessWithEnv``, which aligns with the requirements mentioned in #1373. This function allows spawning processes with custom environment variables, and I have tested it on both Windows and Linux. It works as expected, and below are the screenshots demonstrating its functionality on both platforms.
I will also start working on writing tests, documentation and javascript API.

## Windows
The command executed is echo ``%NEU_VAR%``. For debugging purposes, the output of the command is printed directly. In the actual implementation, the dispatch function is called to handle process events (``stdout``, ``stderr``, ``exit``).

Screenshot_20250202_121011
## Linux
On linux this also works as expected. Hoping this works the same on macOS as well, if needed i can test this on macOS as well.
Screenshot_20250202_115903

## Next steps

I am requesting a review @shalithasuranga on this implementation. Currently, the feature is implemented as a separate function (``spawnProcessWithEnv``). However, it can also be integrated into the existing ``spawnProcess`` function by adding a new parameter (``envList``) for environment variables with minimal changes in the ``spawnProcess`` implementation.

## Here are the two options moving forward:

1. Keep spawnProcessWithEnv as a Separate function
    - If this approach is preferred, I will proceed with writing tests and documentation for the new function.

3. Add a New Parameter (envList) to the Existing spawnProcess Function:
    - This would simplify the API by extending the existing function instead of introducing a new one. If this is the preferred approach, I can refactor the implementation accordingly.

Kindly let me know which direction you would like to take. I am happy to proceed with either option based on your feedback and start working on writing tests, documentation and javascript API.

shalithasuranga:
Hello @SharonIV0x86, thanks so much for sending this pull request. I would like to close this pull request and ask you to submit a new one (if you would like to work on) with the following development considerations:

1. The process library supports environment variables, so just use their implementation without writing additional code to set environment variables
2. Consider using the following parameters:

```js
const command  = 'my-command';
const options = {
  cwd: NL_PATH,
  envs: {
    'VAR1': '1',
    'VAR2': '2' 
  } 
};
os.spawProcess(command, options);
```

(This will increase the Neutralinojs major version since it has a breaking chance, but that's fine -- we can add a note in release notes)

Thanks :tada:

shalithasuranga:
Great! @SharonIV0x86 Make sure to check the status of #1375 before you start developing. Thanks :tada:

PR #1375, opened 2025-01-31
Implement environment variable support for spawnProcess

Oshlok:
Related issue: #1373 

### Description
- This PR adds support for passing custom environment variables and working directories, which wasn't possible with the current `os.spawnProcess` API.

## Changes proposed
- Added `SpawnProcessOptions` struct supporting:
  - `cwd`: Working directory specification
  - `envs`: Custom environment variables
- App developers can simply get the parent process's environment variables using os.getEnvs() if they need:
   `Neutralino.os.spawnProcess('my-command', {envs: {...await Neutralino.os.getEnvs(), 'VAR1', '1', ...}});`
 
**Note:**  This implementation matches Node.js's child_process.exec behavior.

### DIFF
--- a/api/os/os.cpp
+++ b/api/os/os.cpp
@@ -150,11 +150,7 @@ pair<int, int> spawnProcess(string command, const os::SpawnProcessOptions &optio
 
     if (!options.envs.empty()) {
         for (const auto& [key, value] : options.envs) {
-            #if defined(_WIN32)
-            processEnv[helpers::str2wstr(key)] = helpers::str2wstr(value);
-            #else
-            processEnv[key] = value;
-            #endif
+            processEnv[CONVSTR(key)] = CONVSTR(value);
         }
     }
 
--- a/api/os/os.cpp
+++ b/api/os/os.cpp
@@ -133,7 +133,7 @@ os::CommandResult execCommand(string command, const string &input, bool backgrou
     return commandResult;
 }
 
-pair<int, int> spawnProcess(string command, const string &cwd, const map<string, string> &env) {
+pair<int, int> spawnProcess(string command, const os::SpawnProcessOptions &options) {
     #if defined(_WIN32)
     command = "cmd.exe /c \"" + command + "\"";
     #endif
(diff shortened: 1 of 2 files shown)


api/os/os.cpp
```diff
@@ -146,8 +146,40 @@ pair<int, int> spawnProcess(string command, const string &cwd) {
         nextVirtualPid = 0;
     }
 
+    TinyProcessLib::Process::environment_type processEnv;
```
shalithasuranga, line 149: Thanks for sending this pull request @Oshlok. These code changes look good, but I just have one note regarding this section. It seems like you are trying to attach the framework process's environment variables to the new child process. Please note that we always try to implement Node.js API or Electron API behavior to offer a better developer experience through Neutralinojs API. So, I would like to remove this platform-specific code segment and let the process library behave as it is. Then, the `os.spawnProcess(command, options)` will work the same as Node.js's `child_process.exec(command, options)`

App developers can simply get the parent process's environment variables using `os.getEnvs()` if they need:
```js
os.spawnProcess('my-command', {envs: {...await os.getEnvs(), 'VAR1', '1', ...}});
```

We can merge this after modifying the code as suggested above. Thanks :tada:

shalithasuranga (changes requested):

api/os/os.cpp
Oshlok: Yeah sure @shalithasuranga , I think this is much better approach.
I'll modify it accordingly.
Thnx

Oshlok:
@shalithasuranga sir, I have made changes as you said can you please check it.
Also I have created other PRs related to this issue 
- [PR1](https://github.com/neutralinojs/neutralino.js/pull/152)
- [PR2](https://github.com/neutralinojs/neutralinojs.github.io/pull/377)

api/os/os.cpp
```diff
@@ -146,8 +146,20 @@ pair<int, int> spawnProcess(string command, const string &cwd) {
...
+
+    if (!options.envs.empty()) {
+        for (const auto& [key, value] : options.envs) {
+            #if defined(_WIN32)
+            processEnv[helpers::str2wstr(key)] = helpers::str2wstr(value);
```
SharonIV0x86, line 154:
```suggestion
            processEnv[CONVSTR(key)] = helpers::str2wstr(CONVSTR(value));
```
You can utilize these macros to cast strings to appropriate format as per the operating system, as these macros are defined here.

https://github.com/neutralinojs/neutralinojs/blob/4f8c45e3640f8b9ae39b51a862899689a34256e3/helpers.h#L11

shalithasuranga (approved):
Thanks so much @Oshlok for sending this pull request :tada;

PR #1376, opened 2025-02-02
Embedres

tbmale:
## Description

Implement embedding by appending a marker and the resources asar to the executable.

I tried to stay out of the way to the best of my knowledge :)

## Changes proposed

 - Add helper function to calculate resource offset by marker
 - Add ```ResourceModeEmbedded``` and use the offset if this mode is set when retrieving resources

## How to test it

I don't have windows and I didn't dare to go the wine way for msvc, but I tried to do things the cross-platform way.

"It should work"(TM) :)

## Next steps

If you agree, next step would be to mmap the exe, to not open the file at each resource access

## Deploy notes

This PR is accompanied by one for the cli

### DIFF
--- a/buildzri.config.json
+++ b/buildzri.config.json
@@ -28,7 +28,8 @@
             "lib/clip/clip.cpp",
             "lib/clip/image.cpp",
             "lib/infoware/src/**/*.cpp",
-            "lib/efsw/src/efsw/*.cpp"
+            "lib/efsw/src/efsw/*.cpp",
+            "whereami.c"
         ],
         "linux": [
             "lib/tinyprocess/process_unix.cpp",
--- a/helpers.cpp
+++ b/helpers.cpp
@@ -74,24 +74,31 @@ namespace helpers
         int length = wai_getExecutablePath(NULL, 0, NULL);
         char *_exepath = (char *)malloc(length + 1);
         int dirname_length;
+        std::stringstream ss;
         wai_getExecutablePath(_exepath, length, &dirname_length);
         _exepath[length] = '\0';
-        string exepath=std::to_string(*_exepath);
+        string exepath(_exepath);
         long long fsize = fs::getStats(exepath).size;
         fs::FileReaderResult fres = fs::readFile(exepath, {pos : fsize - 8, size : 8});
-        if (fres.status != errors::NE_ST_OK)
+        if (fres.status != errors::NE_ST_OK){
+            debug::log(debug::LogTypeError, errors::makeErrorMsg(fres.status, exepath));
             return {offset:-1,filename:""};
-        long long foffset = 0;
-        for (int i = fres.data.length(); --i >= 0;)
-            foffset = (foffset << 8) | fres.data[i];
-        foffset++;
-        if(foffset<0 || foffset>fsize)
+        }
+        long long foffset;
+        std::memcpy(&foffset,fres.data.data(),sizeof(long long));
+        if(foffset<0 || foffset>fsize){
+            debug::log(debug::LogTypeError, "offset off limits");
             return {offset:-1,filename:""};
-        fres=fs::readFile(exepath, {pos : foffset, size : (long long)marker.length()});
-        if (fres.status != errors::NE_ST_OK)
+        }
+        fres=fs::readFile(exepath, {pos : (long long)foffset, size : (long long)marker.length()});
+        if (fres.status != errors::NE_ST_OK){
+            debug::log(debug::LogTypeError, errors::makeErrorMsg(fres.status, exepath));
             return {offset:-1,filename:""};
-        if(fres.data!=marker)
+        }
+        if(fres.data!=marker){
+            debug::log(debug::LogTypeError, "marker `" +marker+"` not found");
             return {offset:-1,filename:""};
+        }
         return {offset:foffset+marker.length(),filename:exepath};
     }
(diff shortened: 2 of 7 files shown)


shalithasuranga:
Hello @tbmale, thanks so much for your work. I'll keep this pull request open for some time to receive others' feedback on this approach. 

I am not sure about the side effects of injecting binary data into compiled platform-specific binaries. This approach feels like [Steganography](https://en.wikipedia.org/wiki/Steganography). Let's see what other developers think about this.

tbmale:
> Hello @tbmale, thanks so much for your work. I'll keep this pull request open for some time to receive others' feedback on this approach.
> 
> I am not sure about the side effects of injecting binary data into compiled platform-specific binaries. This approach feels like [Steganography](https://en.wikipedia.org/wiki/Steganography). Let's see what other developers think about this.

I know... lazy...

I was thinking of __binary_data_ using something like [postject](https://github.com/nodejs/postject).
That will be somehow semantically better... I think... :)

CosmoMyzrailGorynych:
> I am not sure about the side effects of injecting binary data into compiled platform-specific binaries. This approach feels like [Steganography](https://en.wikipedia.org/wiki/Steganography). Let's see what other developers think about this.

Bun and many other runtimes go the similar way  ¯\\\_(ツ)\_/¯

shalithasuranga:
> > Hello @tbmale, thanks so much for your work. I'll keep this pull request open for some time to receive others' feedback on this approach.
> > I am not sure about the side effects of injecting binary data into compiled platform-specific binaries. This approach feels like [Steganography](https://en.wikipedia.org/wiki/Steganography). Let's see what other developers think about this.
> 
> I know... lazy...
> 
> I was thinking of __binary_data_ using something like [postject](https://github.com/nodejs/postject). That will be somehow semantically better... what do you think?

Impressive solution :tada:. I think this is the most suitable way for Neutralinojs. We can use this library in CLI (i.e., `neu build --embed-res`) and use platform-specific resource loading APIs (i.e., `LoadResource()` on Windows) from the C++ code (Perhaps, we can write the implementation in a separate file, `embeded_resources.cpp` if needed).

Thanks so for posting about `postject` :tada:

CosmoMyzrailGorynych:
> Not sure virus guards are also happy with this approach :)

Antiviruses work based on executable and other files' hashes. If you change an icon, or append data to an executable, the hash changes, too, and thus a base Neutralino executable is completely different for an antivirus from a final executable. What it means in practice is that apps made with Neutralino will need more trust to built up when distributed — unless signed. But, as icons are already replaced, there won't be any difference.

From the app's behavior, antiviruses won't mind an executable reading itself — that's how 90% of installers work anyways, and most of single-executable apps that have images and such inside.

tbmale:
> > > Hello @tbmale, thanks so much for your work. I'll keep this pull request open for some time to receive others' feedback on this approach.
> > > I am not sure about the side effects of injecting binary data into compiled platform-specific binaries. This approach feels like [Steganography](https://en.wikipedia.org/wiki/Steganography). Let's see what other developers think about this.
> > 
> > 
> > I know... lazy...
> > I was thinking of __binary_data_ using something like [postject](https://github.com/nodejs/postject). That will be somehow semantically better... what do you think?
> 
> Impressive solution 🎉. I think this is the most suitable way for Neutralinojs. We can use this library in CLI (i.e., `neu build --embed-res`) and use platform-specific resource loading APIs (i.e., `LoadResource()` on Windows) from the C++ code (Perhaps, we can write the implementation in a separate file, `embeded_resources.cpp` if needed).
> 
> Thanks so for posting about `postject` 🎉
> 
> Edit: seems like https://github.com/lief-project/LIEF can be used for parsing resources. Also, we need to evaluate embedding `resources.neu` vs embedding each resource file I think

Nice finding (LIEF)! Surely more comfortable than implementing my own, platform by platform :)

shalithasuranga:
> . Also, we need to evaluate embedding `resources.neu` vs embedding each resource file I think

I checked this further. Looks like embedding the whole packed `resources.neu` simplifies everything. Some implementation details I imagined:

- Deprecating `--load-dir-res` and introducing `--res-mode=<mode>`, mode -> directory, bundle, embedded (or embedded-bundle)
- Trying to reuse existing resources.neu parsing implementation in `resources.cpp` and `api/res.cpp` 
- Implementing a memory-efficient solution by using LIEF parsed data using streams (looks like LIEF supports streamed data)

Thanks :tada:

IsmaCortGtz:
I'm not sure about how you are working on this (probably postject is a better option), but I think that it could be useful.

Some time ago I was working on a similar feature. The way I solve that (at least in Linux) is to inject the full `resources.neu` next to the neutralino binary, then an 8-byte boundary, and finally, the last 8 bytes, the offset where the resources.neu file starts.

Next, I only modified the resources.cpp to look at itself (using /proc/self/exe in Linux), open an ifstream, and append the search for the last 16 bytes; see if the first 8 of those were the boundary to use the next 8 bytes as an offset when reading the asar file.

I'm sorry if it isn't the best explanation (English isn't my native language). You can see how I deal with different platforms in my fork [here](https://github.com/IsmaCortGtz/neutralinojs/tree/feature/embedded-neu). 

I only tested this on Linux, so I'm not sure if it works on Windows or macOS, but there is code to get the binary path in those OS.

Issue #1292, opened 2024-07-29
Crashes when closing, on Windows, debug enabled.


iparnamaa:
On Windows, Neutralinojs crashes when debug api is enabled and app folder permissions do not allow creating the a log file.

**To Reproduce**
1. Build your app with debug api enabled
2. Delete neutralinojs.log file or .tmp directory from previous run if there
3. Remove Write permission from app directory for current user
4. Execute your app
5. Try to close your app
6. In DevTools there is uncaught (in promise) error, with code NE_RT_NATRTER

**Expected behavior**
App would just close.

**Screenshots/Screen recordings**
![image](https://github.com/user-attachments/assets/639a61e9-91d5-4420-a5eb-5a108f56c94e)
![image](https://github.com/user-attachments/assets/9ba0bf82-2f26-45a7-be0f-7575f7e4fd11)

**Specifications**  
 - OS: Windows
 - Neutralinojs version: 5.3
 - Neutralinojs client library version: 5.3
 - Neutralinojs CLI version: 11.2.2

OmKakatkar:
Can confirm this happens on Windows 11 Version 23H2

SharonIV0x86:
@iparnamaa Kindly take a look at my PR.
https://github.com/neutralinojs/neutralinojs/pull/1379

PR #1379, opened 2025-02-09
Added error hadling for insufficient permissions crash when creating /.tmp directory.

SharonIV0x86:
## Description
As mentioned in https://github.com/neutralinojs/neutralinojs/issues/1292 neutralino app crashes when the folder does not have write permissions. 

The issue most probably arises from here:
https://github.com/neutralinojs/neutralinojs/blob/4f8c45e3640f8b9ae39b51a862899689a34256e3/auth/authbasic.cpp#L40

There is no error handling mechanism when calling ``filesystem::create_directory`` function which most likely causes the neutralino app to crash. If the exception is handled then possibly the app can be saved from crashing, and the debug message is printed on the console.

## Changes proposed
- Added error handling to ``filesystem::create_directory`` function call.

## How to test it
Needs testing on windows, i will be testing this on windows and try to simulate the same environment as mentioned in https://github.com/neutralinojs/neutralinojs/issues/1292  

## Next steps
We need to inspect more possible areas in code which can cause this error to occur, a few that i found are as follows:
- https://github.com/neutralinojs/neutralinojs/commit/5dc121762f8e27315a6ec8fc1dd54d8c84199cd9 (fixed)
- https://github.com/neutralinojs/neutralinojs/blob/4f8c45e3640f8b9ae39b51a862899689a34256e3/api/window/window.cpp#L231
- https://github.com/neutralinojs/neutralinojs/blob/4f8c45e3640f8b9ae39b51a862899689a34256e3/api/storage/storage.cpp#L87

### DIFF
--- a/auth/authbasic.cpp
+++ b/auth/authbasic.cpp
@@ -37,7 +37,12 @@ void init() {
 
 void exportAuthInfo() {
     string tempDirPath = settings::joinAppDataPath("/.tmp");
-    filesystem::create_directories(CONVSTR(tempDirPath));
+    try {
+        filesystem::create_directories(CONVSTR(tempDirPath));
+    } catch (const std::filesystem::filesystem_error& e) {
+        debug::log(debug::LogTypeError, "Failed to create ./tmp directory. Make sure you have sufficient permissions.");
+        debug::log(debug::LogTypeError,std::string(e.what()));
+    }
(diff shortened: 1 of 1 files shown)


shalithasuranga (approved):
Thanks so much for your contribution @SharonIV0x86 :tada:

PR #1389, opened 2025-02-21
:bug: Windows: Fix NE_FS_NOPATHE errors when using server.mount on a path that contains non-ASCII symbols

CosmoMyzrailGorynych:
## Description
Fixes NE_FS_NOPATHE when using `Neutralino.server.mount` with paths that, say, contain cyrillic symbols.

## Changes proposed

 - Use `filesystem::path(CONVSTR(target))` to properly handle paths on Windows
 - Add a test case to ensure the method works with non-latin paths on all systems

## How to test it

 - `npm run test server`

## Next steps
None.

## Deploy notes
None.

### DIFF
--- a/server/router.cpp
+++ b/server/router.cpp
@@ -228,11 +228,12 @@ errors::StatusCode mountPath(string &path, string &target) {
         path = "/";
     }
 
-    const auto targetFsPath = filesystem::path(CONVSTR(target));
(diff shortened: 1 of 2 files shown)


shalithasuranga (approved):
Thanks so much for sending this @CosmoMyzrailGorynych  :tada:

Issue #1274, opened 2024-07-07
JSON utf8 parsing error


OrigamingWasTaken:
**Describe the bug**
Sometimes when given a certain string, neutralino (the app) will crash with this error:
```
libc++abi: terminating due to uncaught exception of type nlohmann::json_abi_v3_11_2::detail::type_error: [json.exception.type_error.316] invalid UTF-8 byte at index 484: 0x2E
```

**To Reproduce**
1. Create a new project
2. Try to split a string that **may** contain invalid characters:
```ts
string.split("\n")
```
3. See error

**Expected behavior**
The app should not crash, instead return an error on the .split() function, or something else to indicate it didn't work.

**Screenshots/Screen recordings**
This is my code, that starts the **Roblox** app, and listens to its output:
```ts
events.on("spawnedProcess", async (evt) => {
			if (evt.detail.id !== instance.id) return;
			switch (evt.detail.action) {
				case "stdOut":
				case "stdErr":
					// Sometimes the string contains invalid utf-8 characters, so we need to fix it.
					// const utf8 = new TextEncoder().encode(evt.detail.data);
					// const validString = new TextDecoder('utf-8', { fatal: false }).decode(utf8);
					// const lines = validString.split("\n");
					const lines = evt.detail.data.split("\n");

					// Process the lines if watchLogs is true
					if (this.watchLogs) {
						this.processLines(lines);
					}
					break;
				case "exit":
					this.emit("exit");
					break;
			}
		});
```

It will crash half of the time on `.split("\n"`

**Specifications**  
 - OS: MacOS 14.5 arm64
 - Neutralinojs version: 5.1.0
 - Neutralinojs client library version: 5.1.0
 - Neutralinojs CLI version: 10.1.0

**Additional context**
*None*

shalithasuranga:
This issue was fixed. Thanks so much for reporting this bug :tada:

PR #1396, opened 2025-02-23
Fix JSON utf8 parsing crash.

SharonIV0x86:
Fixes https://github.com/neutralinojs/neutralinojs/issues/1274
## Description
According to the #1274 The neutralino application crashes due to the uncaught exception of the type error. The error is raised by the ``nlohmann``'s json library when the data to be parsed contains non utf-8 characters. As mentioned in this conversation https://github.com/nlohmann/json/issues/1383. The error is caused when trying to process the ``StdOut`` and ``StdErr`` response generated by the processes from ``execCommand`` and ``spawnProcess`` functions and the processes output can sometimes be complex non utf characters, so there is a need to handle them properly.

The string being sent to this function from here, The ``StdOutHandler`` and ``StdErrHandler``
https://github.com/neutralinojs/neutralinojs/blob/3e4b926860a36a05682beb3d971aa7594155d7cb/api/os/os.cpp#L166

In #1274 The error is caused when listening for ``spawnedProcess`` event and trying to parse the ``evt.detail.data`` this most likely originates from here, when the event is dispatched.
https://github.com/neutralinojs/neutralinojs/blob/3e4b926860a36a05682beb3d971aa7594155d7cb/api/os/os.cpp#L71

more specifically from here, as there is no error handling for checking for invalid data being parsed into a json object.
https://github.com/neutralinojs/neutralinojs/blob/3e4b926860a36a05682beb3d971aa7594155d7cb/api/os/os.cpp#L75

## Changes proposed
- Added a ``sanitizeUTF8`` function in ``helpers.cpp`` and ``helpers.h`` to ensure all output data is valid UTF-8.
- Updated the ``StdOut`` and ``StdErr`` fields of ``commandResult`` to store sanitized data.
- The ``sanitizeUTF8`` function replaces invalid UTF-8 sequences with valid characters. 

## How to test it
While working on this, I realized that the input data could sometimes be really big. To make sure the ``sanitizeUTF8`` function works fast and efficiently, even with large inputs, I ran a benchmark to test its speed and performance. The function handles large amounts of data quickly.

the benchmark for the ``sanitizeUTF8`` function can be seen and executed from here:
https://onecompiler.com/cpp/439wdcew9

## Alternatives
I looked into some of the alternatives for parsing non utf data instead of my own implementation this is what i found.
We could have used well-optimized libraries like 
- [ICU](http://site.icu-project.org/) 
- [Boost.Nowide](https://github.com/boostorg/nowide) for UTF-8 validation and sanitization.

But they might add extra overhead with the external dependencies and ``ICU``(International Components for Unicode) library and boost libraries are not native C++ libraries this idea was dropped. But i am open to incorporating some other header only based unicode parsing libraries. 

## Deploy notes

None.

### DIFF
--- a/api/os/os.cpp
+++ b/api/os/os.cpp
@@ -47,7 +47,6 @@ extern char **environ;
 #include "api/debug/debug.h"
 #include "api/os/os.h"
 #include "api/window/window.h"
-
 #define NEU_MAX_TRAY_MENU_ITEMS 50
 
 using namespace std;
@@ -114,11 +113,15 @@ os::CommandResult execCommand(string command, const os::ChildProcessOptions &opt
     }
 
     auto stdOutHandler = [&](const char *bytes, size_t n) {
-        commandResult.stdOut += string(bytes, n);
+        const string data = string(bytes, n);
+        const string sanitizedStdOut = helpers::sanitizeUTF8(data);
+        commandResult.stdOut += sanitizedStdOut;
     };
 
     auto stdErrHandler = [&](const char *bytes, size_t n) {
-        commandResult.stdErr += string(bytes, n);
+        const string data = string(bytes, n);
+        const string sanitizedStdErr = helpers::sanitizeUTF8(data);
+        commandResult.stdErr += sanitizedStdErr;
     };
 
     if(!options.background && options.envs.empty()) { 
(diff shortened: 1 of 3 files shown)


shalithasuranga:
Thanks so much for sending this pull request @SharonIV0x86. I had to close this since the json library already implements it. Fixed via https://github.com/neutralinojs/neutralinojs/commit/6f7fc8515420205e453084be2c6eef726ccc52e9

SharonIV0x86:
> Thanks so much for sending this pull request @SharonIV0x86. I had to close this since the json library already implements it. Fixed via [6f7fc85](https://github.com/neutralinojs/neutralinojs/commit/6f7fc8515420205e453084be2c6eef726ccc52e9)

Ow that must have escaped my notice, lol. but its better the issue was fixed without introducing much complexity in the code. 🥳

PR #1399, opened 2025-02-25
Automatically log extensions' stdOut, stdErr, and exit

Notebooked:
## Description

This PR modifies the extensions loader in order to properly log the console output/error of the extension process. Any console.log or errors thrown inside of an extension process will have the output logged through Neutralinojs's debug logging system.

This is in response to my experiences with struggling to debug extensions as well as [this issue](https://github.com/neutralinojs/neutralinojs-cli/issues/189), though this is not an issue with the CLI.

## Changes proposed

 - Modify the extensions loader to use os::spawnProcess instead of os::execCommand so that the extension command can be run in the background and have the output be asynchronously logged using the Neutralinojs debug logger.
 - Add the ability for os::spawnProcess to run passed-in handler functions for stdOut, stdErr, and process exit (internally, not in the controller).

## How to test it

 - Enable extensions in the config
 - Try to run the sample extension with the log calls
 - Optionally uncomment the error lines in the sample extension
 - Run the client

## Notes

It is possible that adding a new argument configuration for os::spawnProcess would be too intrusive, though I believe it is not. This was only done so that the callback functions could be asynchronously called, but an alternative would be to connect to Neutralino's events, though I do not see how this is possible or necessary.

Some ideas for changes could be to include the extension ID with the console message (complete), and to include a setting that turns this logging on or off.

## Request for Feedback

I am not too familiar with C++ syntax or common practices, so if I violated any, I would be happy to change my code. I truly appreciate any feedback, suggestions, or corrections from the community.


shalithasuranga:
Hello, thanks so much for your work. I had to close this since https://github.com/neutralinojs/neutralinojs/commit/9ba940d5b6d3ce17637f91bf0b6dc02218a2d827 implemented this without creating a new function and modifying the extension source. We hope you'll do another contribution in the future :tada:

PR #1419, opened 2025-05-28
feat: add native dragging operation (for usage with the SDK)

jay-taelien:
## Description
Adds the window.beginDrag method, allowing for custom titlebars to utilize the native platform APIs when dragging a window. Pairs with the new implementation of setDraggableRegion in the SDK PR (coming soon), allowing for matched behavior when utilizing.
**NOTE:** PLEASE TEST THE APPLE/LINUX IMPLEMENTATION. I've only tested Windows and can confirm everything works as expected. The Apple/Linux code *should* work, but I don't have the right devices to test.

## Changes proposed
 - Add window.beginDrag as a receivable WS event to utilize native window dragging

## How to test it
- Send window.beginDrag over the websocket with the x and y of the mouse position. Moving your mouse around will subsequently move the window. **For a less obtuse test**, pair this with the Neutralino JS SDK with the setDraggableRegion PR I'm about to send to see it in action.

Examples to test with the new SDK function:
```js
// Register a region and exclude a button by ID
const region = await setDraggableRegion('my-draggable-div', {  // alternatively: document.getElementById('my-draggable-div')
  exclude: ['initial-excluded-button'], // alternatively:  exclude: [document.getElementById('initial-excluded-button')]
})

// Exclude additional elements using DOM references or IDs
region.exclusions.add('another-excluded-button')                // ✅ Works
region.exclusions.add(document.getElementById('input-field'))   // ✅ Works
region.exclusions.add(['el1', document.getElementById('el2')])  // ✅ Works
region.exclusions.add('one', ['two', 'three'])                  // ✅ Works (mixed styles)

// Remove a previously excluded element
region.exclusions.remove('another-excluded-button')             // ✅ Works
region.exclusions.remove(['initial-excluded-button'])           // ✅ Works
region.exclusions.remove('input-field', 'nonexistent')          // ✅ Safe: no error on non-registered

// Clear all exclusions
region.exclusions.removeAll()                                   // ✅ Works

// Clear the draggable region
await unsetDraggableRegion('my-draggable-div')

// ❌ Attempt to use exclusions after region has been unset
region.exclusions.add('new-button')                             // ❌ Throws: NE_WD_NOTDRRE
region.exclusions.remove('some-button')                         // ❌ Throws: NE_WD_NOTDRRE
region.exclusions.removeAll()                                   // ❌ Throws: NE_WD_NOTDRRE
```

 - All current spec tests for `window` still pass with this implementation.

## Next steps
PR is incoming to the Neutralino.js JS SDK to utilize this new method of setting a draggable region!

## Deploy notes
None.

### DIFF
--- a/api/window/window.cpp
+++ b/api/window/window.cpp
@@ -727,6 +727,34 @@ void move(int x, int y) {
     #endif
 }
 
+void beginDragNative(int x, int y, int button) {
+
+#if defined(__linux__) || defined(__FreeBSD__)
+    gtk_window_begin_move_drag(GTK_WINDOW(windowHandle),
+        button, x, y, GDK_CURRENT_TIME);
+
+#elif defined(_WIN32)
+    ReleaseCapture();
+    SendMessage(windowHandle, WM_SYSCOMMAND, SC_MOVE | HTCAPTION, 0);
+
+#elif defined(__APPLE__)
+    NSEvent* ev = ((id(*)(id, SEL, NSUInteger, CGPoint, NSUInteger,
+        NSTimeInterval, NSInteger, BOOL, NSInteger, CGFloat))
+        objc_msgSend)("NSEvent"_cls,
+            "mouseEventWithType:location:"
+            "modifierFlags:timestamp:windowNumber:"
+            "context:eventNumber:clickCount:pressure:"_sel,
+            1 /*NSLeftMouseDown*/,
+            CGPointMake(x, y),
+            0, 0,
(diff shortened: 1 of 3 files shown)


jay-taelien:
This also allows native Windows Snap to active when dragging the Neutralino window to the edge of the screen, which most modern desktop app frameworks already have out of the box.

There's still a bug when making a borderless window and then setting resizable to true with setSize() wherein the top of the window has a glaring dark bar, so if you find the root cause of that and fix that somehow that would fully round out the borderless window feature set (at least on Windows)

PR #1459, opened 2025-08-28
Postject embedded resources

IsmaCortGtz:
## Description

Embed the `resources.neu` file inside the pre-built binary using [`node/postject`](https://github.com/nodejs/postject).

## Changes proposed

 - Add the official [`postject-api.h`](https://github.com/nodejs/postject/blob/main/postject-api.h) to handle pointers efficiently like `NODE SEA`, reusing the ASAR format for the resources file.
 - Add a dummy resource file for Windows inside `lib/postject/windows_dummy.res`. This is needed by `postject` to know where to add the `resources.neu` file.
 - Add a new `--res-mode` flag as suggested in #1376. Accepting the values `directory`, `bundle` and `embedded`.

***Note:*** By default the app tries with `embedded` mode, then with `bundle` and `directory` respectively, unless the value is overridden using a flag.

## How to test it

Following the [`nodejs/postject`](https://github.com/nodejs/postject) usage example, you can embed a `resource.neu` file. Use `NEUTRALINOJS_RESOURCES_NEU` as the resource name.

```bash
# Inject using CLI
npx postject neutralino-linux_x64 NEUTRALINOJS_RESOURCES_NEU resources.neu
```

```js
// Inject using code
import { inject } from 'postject';
import fs from 'node:fs';

const content = fs.readFileSync('./resources.neu');
await inject('./neutralino-linux_x64', 'NEUTRALINOJS_RESOURCES_NEU', content);
```

Then, you can run the neutralino as usual.

```bash
# Embedded mode
./neutralino-linux_x64

# Bundle mode (will search a resource.neu)
./neutralino-linux_x64 --res-mode=bundle

# Directory mode
./neutralino-linux_x64 --res-mode=directory
./neutralino-linux_x64 --load-dir-res
```

## Next steps

Implement an embed cli flag/command at [`neutralinojs/neutralinojs-cli`](https://github.com/neutralinojs/neutralinojs-cli/). Maybe something like:

```bash
# Delete resource.neu from dist folder after embed
neu build --embedded
```

## Deploy notes

**IMPORTANT**

From linux x64 I can embed any linux arch, win_x64 and mac_x64, but postject fails with mac_arm64 (and universal).

With arm64 looks like an arch-incompatibility problem.

Universal builds internally stores both archs, so the FUSEL (the way how postject detects if a resource si embedded) is duplicated, additionally, the universal binary has arm64 bytecode so it will fail for this reason too.

The solutions seems to be embed both binaries from `macos` (I cannot test that) or maybe a docker arm64 image could be enogh, but I don't know if this is a good solutions for cross-platform development.


shalithasuranga:
Hello @IsmaCortGtz, Thanks so much for sending this pull request with a detailed explanation. The code modifications look clean and great -- we can add this to the codebase. I will test with the postject cli too and merge. Thanks :tada:

Issue #1426, opened 2025-06-17
Net API Proposal


IsmaCortGtz:
# Net API Proposal for `neutralinojs`

I'm quite interested in the Net API in `neutralinojs`, so I've been working on examples and reading about possible implementations. I've collected the following options (*some simpler than others*). I've tried to gather as much information as possible about each one to see potential advantages, disadvantages, and how they would impact the building process.

Before listing the options, I'd like to emphasize that none of them is a perfect solution. Handling HTTP requests, and specifically HTTPS, requires a lot of code to properly manage SSL/TLS certificates, in addition to implementing different standards and algorithms that are not feasible to replicate manually.

# Option 1: `libcurl` with dynamic linking

This is probably the best alternative in terms of performance, size, and compilation flow, although it has its drawbacks, as already mentioned in PR #1064.

#### macOS

`libcurl` is included in the macOS core, so there's no downside to dynamically including libcurl (it will already handle SSL and any other dependencies without problems).

#### Windows

This is probably the most difficult point. Although some versions of Windows include `curl.exe`, it doesn't include what's necessary for use in C++, so static linking isn't an option here.

#### Linux

This platform is quite problematic. Although most distributions already include this library and `libcurl` by default, an advanced user could work their way up to a system without it. Despite this, I honestly believe it's still the best option, as currently Neutralino isn't 100% portable since it depends on libraries like `gtk` and `webkitgtk`, so adding `libcurl` makes perfect sense. This way, we could let the specific build for that system handle the certificates, which makes things easier.

# Option 2: `libcurl` with static linking

#### macOS

Again, macOS already includes the library for dynamic linking by default, `/usr/lib/libcurl.dylib`, so making a static link would only unnecessarily increase the build size, but it is possible.

If you prefer this option, we could compile `libcurl` using `secure-store`. This way, we take advantage of the system certificates and avoid compiling an SSL backend.

#### Windows

This is probably the best alternative on Windows, as we could even use an official precompiled `.dll` and avoid manual compilation. However, if necessary, it is possible to compile it with `schannel`. This way, we avoid any other SSL backend and do not increase the final size even further.

#### Linux

This alternative is viable and a good idea (although, personally, I think dynamic link is better). It is possible to compile `libcurl` + `mbedtls` statically (*it is possible to use `openssl`, but the size increase is too much to be worth it*), and the final Linux binary only increases by `2 MB` compared to the original size. You can see the results in PR #1064.

[...]

PR #1465, opened 2025-09-22
Build using rsp files

IsmaCortGtz:
## Description
Update `scripts/bz.py` to use a [`rsp` file](https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2023-0/use-response-files.html) and avoid the [CMD string limitation](https://learn.microsoft.com/en-us/troubleshoot/windows-client/shell-experience/command-line-string-limitation).

Windows has a limitation of 8,191 characters. I found this problem working on the `Net API` with `bearssl` (#1426).
BearSSL has a very modular architecture, which implies a lot of source files; the result is around 9,960 characters.

MacOS and Linux have a bigger limit, so the problem exists only with Windows, but I think it is better to keep the process consistent across platforms.

## Changes proposed

 - Update `scripts/bz.py` to generate and use `buildzri.rsp`
 - Update `.gitignore` to ignore `*-rsp`

## How to test it

 - Just compile it using `./scripts/bz.py --verbose`. You should use the verbose flag to see what is split in the `rsp` file and what keeps in the CLI.


IsmaCortGtz:
I will close this PR and move it to the [`codezri/buildzri`](https://github.com/codezri/buildzri) repository.

Issue #1466, opened 2025-09-25
Faster builds for contributing to neutralinojs framework


garygreen:
Currently the build tool used to compile Neutralino is [**_BuildZri_**](https://codezri.org/docs/buildzri/intro/) and a Python script - which has a few problems:

❌ Although it has some documentation, it's not well known in the ecosystem, hard to reason about
❌ It's slow to rebuild. Every change made needs a full compilation, which slows development

Would you be open to a PR to migrate to a new build process, of using CMake and Ninja? These would provide immediate benefits:

✅ **Well known, cross platform**, industry standard, lightweight tools
✅ **Fast**. Incremental rebuilds nearly instantaneous, used by likes of Google to build Chromium

I have a "proof of concept" version working locally, just using basic Makefile process and rebuilds is near instantaneous. I'm new to C++, so I'm learning as I go and apologise if any of my understanding in regards to development pipeline is off, but it's a better much development experience when rebuilds are fast 😀

PR #1497, opened 2025-12-17
Migrate to CMake build

IsmaCortGtz:
## Description
As mentioned in #1466, #1426, `buildzri` is a very cool tool for lightweight multi-platform projects, but at the current state of `Neutralino`, the tool is not enough, and the build time is very long. For this (_and others listed below_), this PR aims to migrate to `CMake` and `Ninja`.

## Changes proposed
 - Replace `buildzri` with `CMake`.
 - Add CMake toolchains for Linux ARM build to speed up compilation time.

### Dependencies

- `CMake` 3.16 or higher.
- `Ninja`

### Building

1. Create a `build` folder.

```bash
mkdir -p build
```

2.  Generate `Ninja` files.

```bash
# Current OS and arch
cmake -DCMAKE_BUILD_TYPE=Release . -G Ninja -B build

# Or Custom toolchain (for cross-compile)
cmake -DCMAKE_BUILD_TYPE=Release . -G Ninja -B build -DCMAKE_TOOLCHAIN_FILE=cmake/toolchains/linux-armhf.cmake
```

3. Build using `Ninja` (_you can use `make` if you like, but it's slower_).

```bash
ninja -C build -j$(nproc)
```

The result will be in the `bin/` directory as usual. You can securely delete the `build/` directory after finishing.

## Next steps

Update the GitHub Actions and remove `buildzri` configuration files. I already made some tests at [`this fork`](https://github.com/IsmaCortGtz/neutralinojs/tree/feature/CMake-old), so there shouldn't be a problem.

## Advantages

There are probably more, but I would like to highlight the following two.

### Dependencies compatibility

I think this is the most important one. Apps using buildzri are compiled with only one `g++` (_or any other compiler_) call, so every source file is compiled as the same language and linked at once. 

But the big advantage of modern `C/C++` is that you can compile different languages, build them in steps, and link them together in the same binary using tools like `CMake`.  

I won't be very detailed in this point. You can see #1426 for more details, but I'm currently working in the `Net API`. At first I managed to compile `bearssl` using `buildzri` (_any other library like `mbedtls` is impossible to build with `buildzri`_), but `bearssl` is a pretty old implementation, and some websites won't work for a certificate validation problem, so here the solution is to migrate to `CMake` and `mbedtls`.

This was only the case that I got, but in the future the Neutralino API would grow in many directions, and the ability to use `C` or even `Rust` dependencies would give the project a bigger number of options.

### Build time

The build time was reduced in a very heavy way. Running on my local machine (_Debian 12 - Intel i5 11th_), these were the results:

#### `BuildZri`

```bash
#!/bin/bash
python3 ./scripts/bz.py

# Results using 'time ./build.sh'
real    1m40.225s
user    1m34.976s
sys     0m4.674s
```

#### `CMake`

```bash
#!/bin/bash
mkdir -p build
cmake -DCMAKE_BUILD_TYPE=Release . -G Ninja -B build
ninja -C build -j$(nproc)

# Results using 'time ./build.sh'
real    0m34.815s
user    2m57.331s
sys     0m8.851s
```

I tested in GitHub Actions too (_you can see more detailed logs [`here`](https://github.com/IsmaCortGtz/neutralinojs/actions)_). These were the results (this result represents the whole runner time, which includes dependencies, install, and startup):

| **OS-arch**          | [**BuildZri**](https://github.com/IsmaCortGtz/neutralinojs/actions/runs/19604916625) | [**CMake**](https://github.com/IsmaCortGtz/neutralinojs/actions/runs/18884299671) |
|----------------------|-----------|--------------|
| **linux-x64**        | 2m 55s    | 1m 30s       |
| **linux-arm64**      | 26m 43s   | 3m 27s       |
| **linux-armhf**      | 26m 54s   | 2m 45s       |
| **windows-x64**      | 1m 54s    | 1m 51s       |
| **darwin-x64-arm64** | 4m 58s    | 2m 27s       |

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,10 @@
 /bin/neutralino-*
 /bin/resources.neu
 /bin/resources/js/neutralino.*
+/build
+/build-*
+/build-arm64
+/build-armhf
 
 # Platform specific 
 Thumbs.db
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,253 @@
+cmake_minimum_required(VERSION 3.16)
+project(NeutralinoJS LANGUAGES C CXX)
+set(NEU_VERSION "6.3.0")
+set(NEU_CXX_STANDARD 17)
+
+execute_process(
+  COMMAND git rev-parse HEAD
+  WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+  OUTPUT_VARIABLE NEU_COMMIT
+  OUTPUT_STRIP_TRAILING_WHITESPACE
+  ERROR_QUIET
+)
+
+# =========================
+# Source files
+# =========================
+file (GLOB SOURCES *.cpp)
+file(GLOB_RECURSE RECURSIVE_SOURCES
+  auth/*.cpp
+  server/*.cpp
+  api/*.cpp
+  lib/tinyprocess/process.cpp
+  lib/easylogging/easylogging++.cc
+  lib/platformfolders/platform_folders.cpp
+  lib/clip/clip.cpp
+  lib/clip/image.cpp
+  lib/infoware/src/**/*.cpp
+  lib/efsw/src/efsw/*.cpp
+)
+list(APPEND SOURCES ${RECURSIVE_SOURCES})
+
+if(WIN32)
+  set(NEU_OS_NAME "win")
+  file(GLOB_RECURSE OS_SOURCES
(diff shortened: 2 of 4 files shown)


shalithasuranga (approved):
Thanks so much @IsmaCortGtz for sending this pull request with a detailed description. Looking forward to seeing your GitHub Actions-related updates.. :tada:

PR #1506, opened 2026-01-04
Fix query parameters being dropped in Neutralino.window.create

### LINKED ISSUE #1480: Neutralino.window.createdrops URL search parameters after the first one
**Describe the bug**
When using Neutralino.window.create()with a URL containing multiple search parameters, only the first parameter is preserved while others are dropped.

**To Reproduce**
1. Call
    Neutralino.window.create("https://www.demo.com/query?a=b&c=d&e=f");
2. Observe that the opened window only receives ?a=b
    (Expected: full query string ?a=b&c=d&e=fshould be preserved)

**Specifications**  
 - OS: [e.g. Windows 10 x64]
 - Neutralinojs version: [e.g. v6.3.0]
 - Neutralinojs client library version: [e.g. 6.3.0]

Dh4nu5hwebdev:
### Problem
When creating a window with a relative URL that contains search parameters (e.g. `/test.html?a=b&c=d`), the opened window loses the query string and `window.location.search` becomes empty.

### Root cause
The internal HTTP server routed requests using the full resource string returned by the connection, which caused the query string to be discarded when serving the file.

### Fix
This change separates the request path from the query string when routing the resource, while allowing the browser to retain the full URL (including search parameters).

### Result
- Multiple query parameters are preserved correctly
- No behavior change for URLs without query strings
- Minimal and localized fix

Fixes #1480

### DIFF
--- a/server/neuserver.cpp
+++ b/server/neuserver.cpp
@@ -212,6 +212,7 @@ void handleHTTP(websocketpp::connection_hdl handler) {
     con->set_body(routerResponse.data);
     con->replace_header("Content-Type", routerResponse.contentType);
 
+
     if(applyConfigHeaders) {
         __applyConfigHeaders(con);
     }


shalithasuranga:
Query parameters are already ignored (see: https://github.com/neutralinojs/neutralinojs/blob/228fb74fbda86be5701997a56a9b1148c1df6451/server/router.cpp#L407). I think the issue describes a different problem.  Thanks for sending this pull request :)

Issue #1493, opened 2025-12-09
Draggable region divs block click events


ian-speckart:
**Describe the bug**
Click and double-click events aren't triggered on elements that are made draggable by setDraggableRegion().
This occurs to event listeners added on the document or on the draggable element.

**To Reproduce**
1. Create a new Neutralino project
2. Add the permission to drag the app's window (in neutralino.config.json, in the nativeAllowList array, add "window.*")
3. Add code to make a region draggable and to listen for clicks:
```
Neutralino.window.setDraggableRegion("neutralinoapp");

document.addEventListener("click", function (event) {
  console.log("click");
});
```
4. Click on the draggable region and notice how the click events aren't triggered. You can use 'neutralinoapp' or the 'info' div to test.

**Expected behavior**
The clicks on the draggable region should be detected, or the documentation should inform about this limitation.

**Specifications**  
 - OS: Windows 11 Pro (no WSL)
 - Neutralinojs version: v6.4.0
 - Neutralinojs client library version: v6.4.0
 - Neutralinojs CLI version: v11.6.0

**Additional comments**
This started happening with the new window dragging system. The bug is not present in v5.5.0.

yug105:
I was able to reproduce this on macOS (M2), but the behavior is different from Windows.

On macOS, I can both drag the window and still receive click events on the draggable region created via setDraggableRegion(). I’ve attached a short video demonstrating this.

This suggests that the behavior is OS-level dependent:

On Windows, draggable regions appear to consume mouse events before they reach the webview

On macOS, mouse events may still propagate to JavaScript

This might not be a bug in Neutralinojs itself, but rather an OS-specific limitation. It may be worth documenting that setDraggableRegion() behaves differently across platforms, especially for Windows users.

https://github.com/user-attachments/assets/0cb5167a-4837-4fa5-a039-718683e701e2

PR #1509, opened 2026-01-07
 ##1493 fix(windows): resolve click events being blocked in draggable regions

subhu339:
#1493
## Description
On Windows, SendMessage with SC_MOVE starts a modal loop that blocks the message queue immediately. This makes it impossible for the WebView to fire click or mouseup events because the drag starts before the click is registered.
I've switched this to PostMessage. This lets the OS queue the drag command so that the JS events can finish processing first

## Changes proposed
Replaced SendMessage with PostMessage in beginDragNative.
Added GetCursorPos check before releasing capture.

## How to test it
Set a div as a draggable region.
Add an onclick listener to that div.
Click it—it should now move the window AND trigger the JS alert/log.

## Next steps
None.

## Deploy notes
None.

### DIFF
--- a/api/window/window.cpp
+++ b/api/window/window.cpp
@@ -903,22 +903,24 @@ void move(int x, int y) {
 }
 
 void beginDragNative() {
-
-    #if defined(__linux__) || defined(__FreeBSD__)
+#if defined(__linux__) || defined(__FreeBSD__)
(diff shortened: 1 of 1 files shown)


shalithasuranga (approved):
Thanks so much for sending this :)

PR #1510, opened 2026-01-11
bug fixed #1470: Linux: File Dialog fails to open on systems with xdg…

### LINKED ISSUE #1470: Linux: File Dialog fails to open on systems with xdg-desktop-portal (e.g., modern Arch Linux)
#### Bug Description

After a recent system update on Arch Linux, the file dialog (`Neutralino.os.dialog.openFile`) no longer opens. The call fails silently, with no error messages in the NeutralinoJS application log or the system console.

This issue is caused by the mandatory installation of `xdg-desktop-portal` as a dependency for recent versions of `gtk4`. The underlying WebKitGTK webview appears to default to using this portal for file dialogs, which can fail depending on the user's desktop environment or window manager configuration.

Crucially, the standard GTK workaround of launching the application with the `GTK_USE_PORTAL=0` environment variable **does not work**, suggesting the issue needs to be addressed within the NeutralinoJS/webview initialization layer.

#### Steps to Reproduce

1.  Use a modern Arch Linux system (or another Linux distribution where `gtk4` has a hard dependency on `xdg-desktop-portal-gtk`).
2.  Ensure `gtk4` and its dependency `xdg-desktop-portal-gtk` are installed and up-to-date.
3.  Create a basic NeutralinoJS application with a function that calls `await Neutralino.os.dialog.openFile({ title: 'Open a file' });`.
4.  Build and run the application.
5.  Trigger the function to open the file dialog.

#### Expected Behavior

A native GTK file selection dialog should appear, allowing the user to select a file.

#### Actual Behavior

Nothing happens. The dialog does not appear. The application does not freeze or crash, and the JavaScript promise from the `dialog.openFile` call does not seem to resolve.

#### System Details

*   **OS:** Arch Linux (Rolling Release)
*   **Neutralinojs Version:** tested with `6.2.0` and `6.3.0`
*   **Kernel Version (at time of issue):** `6.16.10.arch1-1`
*   **Key Package Versions (that introduced the issue):**
    *   `gtk4`: `1:4.20.2-2`
    *   `xdg-desktop-portal`: `1.20.3-2`
    *   `xdg-desktop-portal-gtk`: `1.15.3-1`

#### Workaround

The only effective workaround is to remove the portal system entirely, which requires downgrading `gtk4`.

1.  Downgrade the `gtk4` package to a version that does not have a hard dependency on `xdg-desktop-portal-gtk` (e.g., using the Arch Linux Archive).
2.  Remove the `xdg-desktop-portal` and `xdg-desktop-portal-gtk` packages.
    ```bash
    sudo pacman -Rns xdg-desktop-portal xdg-desktop-portal-gtk
    ```
3.  Add `gtk4` to the `IgnorePkg` list in `/etc/pacman.conf` to prevent the issue from reoccurring on the next system update.

After these steps, the NeutralinoJS file dialog works correctly again.

#### Proposed Solution / Discussion

Since `GTK_USE_PORTAL=0` is ineffective, a more robust solution within NeutralinoJS is needed. Is it possible to programmatically disable the use of the portal system when creating the GTK webview instance?

This would make NeutralinoJS applications more resilient to system configuration changes on Linux, which is a common challenge on rolling-release distributions like Arch. It would ensure that the reliable, direct GTK file chooser is always used.

Thank you for your work on this great project

Saurabh-1785:
## Description

Fixes file dialogs (`os.showOpenDialog`, `os.showFolderDialog`, `os.showSaveDialog`) failing silently on modern Linux distributions where GTK4 has a hard dependency on `xdg-desktop-portal-gtk` (e.g., Arch Linux).

## Changes proposed

- Added `setenv("GTK_USE_PORTAL", "0", 1)` before `execvp()` in `lib/filedialogs/portable-file-dialogs.h` to disable GTK portal usage in spawned zenity/dialog processes
- This ensures direct GTK file chooser dialogs are used instead of the potentially misconfigured portal system

## How to test it

- Build NeutralinoJS on a Linux system with `xdg-desktop-portal-gtk` installed
- Run a NeutralinoJS app that calls `Neutralino.os.showOpenDialog()`
- Verify the file dialog appears correctly (previously it would hang silently)
- Safe for all Linux configurations: no-op on kdialog (Qt), ignored on systems without xdg-desktop-portal

## Git Commands

Run these in your WSL terminal:

```bash
cd ~/neutralinojs

# Create a new branch
git checkout -b fix/linux-file-dialog-xdg-portal

# Stage the change
git add lib/filedialogs/portable-file-dialogs.h

# Commit
git commit -m "fix(linux): prevent file dialog hang on systems with xdg-desktop-portal"

# Push (replace 'origin' with your fork remote if needed)
git push -u origin fix/linux-file-dialog-xdg-portal
```

---

## Next steps

None.

## Deploy notes

None.
```


shalithasuranga (changes requested):

shalithasuranga:
Closed until this receives requested updates. Thanks :tada:

Issue #1462, opened 2025-09-17
Suggestion: Use logical pixels for window dimensions to respect screen scaling (DPI)


viocha:
**Is your feature request related to a problem? Please describe.**

Currently, the window `width` and `height` in `neutralino.config.json` are specified in physical pixels.

```json
  "modes": {
    "window": {
      "title": "test-app",
      "width": 2400,
      "height": 1600
    }
  }
```

This approach does not account for the system's display scaling factor (or `devicePixelRatio` on the web), leading to inconsistent window sizes and a poor user experience across devices with different screen densities.

A window that is correctly sized on a high-DPI display will appear oversized or even overflow the screen on a standard-DPI display.

**A Concrete Example:**

Let's consider two different devices:

**Device A (High-DPI / 4K Laptop):**
*   **Physical Resolution:** 3840 x 2400
*   **System Scaling:** 300%
*   **`devicePixelRatio` in frontend:** `3`
*   **Logical Resolution (what the user sees and what CSS/JS uses):** 1280 x 800 (`3840/3` x `2400/3`)

**Device B (Standard-DPI / FHD Monitor):**
*   **Physical Resolution:** 1280 x 800
*   **System Scaling:** 100%
*   **`devicePixelRatio` in frontend:** `1`
*   **Logical Resolution:** 1280 x 800

**The Problem:**

To create a reasonably sized window on **Device A**, I might set the configuration to `width: 2400`, `height: 1600`. This looks good on the 3840x2400 physical screen.

However, when I run the same application with the same configuration on **Device B**, the window (2400px wide) is much larger than the screen's physical width (1280px), causing it to overflow and become unusable.

The core issue is that developers have to think in terms of physical pixels, which they are often unaware of, instead of the logical pixels they work with every day in web development.

**Describe the solution you'd like**

I suggest that the `width` and `height` values in the configuration be interpreted as **logical pixels** (also known as device-independent pixels).

Neutralino should then use the system's scaling factor to calculate the appropriate physical pixel dimensions when creating the window.

For example, if the configuration is set to:
```json
"width": 900,
"height": 600
```

The resulting physical window size would be:
*   On **Device A** (300% scaling): `900 * 3` x `600 * 3` = **2700 x 1800** physical pixels.
*   On **Device B** (100% scaling): `900 * 1` x `600 * 1` = **900 x 600** physical pixels.

In both cases, the window would occupy a consistent, predictable portion of the usable screen space, providing a consistent experience for all users.

[...]

PR #1514, opened 2026-01-14
Added DPI-aware logical window sizing with optional pixelUnits

prasu-baran:
## Description
Neutralino currently interprets `window.width` and `window.height` as physical pixels, which leads to inconsistent window sizes across displays with different DPI scaling. This PR introduces an opt-in logical pixel model, aligning Neutralino with modern frameworks like Electron and making window sizing predictable across devices.

## Changes proposed
- Add optional `pixelUnits: "logical"` to window options.
- When enabled, convert logical sizes to physical pixels on window creation and resize.
- Make `getSize()` return logical pixels

## How to test it
1.Run the app on systems with different DPI scaling (100%, 125%, 200%+).
2.Verify the window appears consistently sized.
3. In DevTools, run: await Neutralino.window.getSize()

#1462

### DIFF
--- a/api/window/window.cpp
+++ b/api/window/window.cpp
@@ -189,28 +189,6 @@ bool __isFakeHidden() {
 	}
 	return false;
 }
-#endif
-
-double __getScaleFactor() {
-	#if defined(_WIN32)
-    return GetDpiForSystem() / 96.0;
-
-	#elif defined(__APPLE__)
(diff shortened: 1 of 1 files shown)


shalithasuranga (approved):
Thanks @prasu-baran. This is a nice feature to have for Neutralinojs :tada:

PR #1519, opened 2026-01-16
fixed(#1518) Fix extension handling for zenity

### LINKED ISSUE #1518: showOpenDialog hangs when pfd uses zenity
**Describe the bug**

On some Linux systems, `showOpenDialog` / `showSaveDialog` (via portable-file-dialogs → zenity) hangs when file filters are provided.

The underlying `zenity` invocation receives `--file-filter` arguments with **trailing whitespace in the glob pattern** (e.g. `"*.cmproj "`). On Manjaro (zenity 4.2.1), this is invalid syntax and causes zenity to hang silently: no dialog is shown, the process does not exit, and the JS promise is never resolved.

**To Reproduce**

1. Call `showOpenDialog` or `showSaveDialog` with file filters.
2. On Manjaro (XFCE), no dialog appears and the call never resolves.

**Expected behavior**

A file dialog should be displayed and the promise should resolve.

**Environment**

- OS: Manjaro Linux (XFCE)
- Neutralino CLI: v11.6.0
- Neutralino binaries/client: v6.4.0
- zenity: 4.2.1

**Additional context / investigation**

This can be reproduced directly with zenity:

```bash
zenity --file-selection --save --file-filter "Project Files|*.cmproj"
```
✔ Dialog appears
```bash
zenity --file-selection --save --file-filter "Project Files|*.cmproj "
```
✖ No dialog appears; zenity hangs and does not exit

When invoking showSaveDialog, the trailing whitespace can be observed in the arguments passed to zenity (verified via a PATH shim).

This suggests the whitespace is introduced while constructing the filter string before calling portable-file-dialogs.

One likely source is the filter construction in api/os/os.cpp (around line 283):

https://github.com/neutralinojs/neutralinojs/blob/3dd12c34b15a8b2d456dd57ec9d32d56f0f6780e/api/os/os.cpp#L283

If filter patterns were joined without a trailing space (or trimmed before passing to portable-file-dialogs), the issue may be resolved.

It is possible this could also be considered a downstream zenity or portable-file-dialogs robustness issue; I leave that determination to the maintainers.

matortheeternal:
This changes the` __extensionsToVector` helper function to not produce a trailing space on the last extension, which succeeds in some environments but causes a catastrophic failure in zenity (no dialog, process hangs).

This resolves #1518 

## How to test it
On Manjaro XFCE or another linux operating system with zenity 4.2.1 installed, invoke `showOpenDialog` with one or more filter arguments.

Tests of the dialog and other dialogs on other operating system should also be performed to verify this change did not introduce any issues (unlikely, but can never be too sure).

I built neutralinojs and ran it locally to verify it on my system.

## Next steps
None.

## Deploy notes
None.

### DIFF
--- a/api/os/os.cpp
+++ b/api/os/os.cpp
@@ -280,7 +280,7 @@ vector<string> __extensionsToVector(const json &filters) {
         filtersV.push_back(filter["name"].get<string>());
 		const auto &exts = filter["extensions"];
         string extensions = "";
-        for (auto i = 0; i < exts.size(); ++i) {
+        for (int i = 0; i < exts.size(); i++) {
             extensions += "*." + exts[i].get<string>();
 			if (i + 1 < exts.size()) {
 				extensions += " ";


shalithasuranga (approved):
Thanks so much for your contribution :tada: @matortheeternal

Issue #1377, opened 2025-02-03
Handle HTTP 206 Partial Content requests to allow video seeking/streaming


stolksdorf:
**Is your feature request related to a problem? Please describe.**
Attempting to jump to a specific timestamp on a `<video>` tag with a source served via `Neutralino.server.mount()` jumps you back to the start of the video. 

**Describe the solution you'd like**
Implement HTTP 206 Partial Content responses for the `server/neuserver.cpp` if the request headers have a specified range. I'm pretty sure only implementing [Single Part](https://www.rfc-editor.org/rfc/rfc9110.html#name-single-part) handling will be sufficient for video tag streaming. 

**Describe alternatives you've considered**
The only workaround I've found is to implement my own streaming file server (node.js + read streams) and run it alongside the Neutralino server.

**Additional context**
- [Mozilla HTTP 206](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/206)

yashsaraswat2004:
Can I work on this issue?

PR #1520, opened 2026-01-18
Add HTTP 206 Partial Content support for Range requests (video streaming)

prasu-baran:
## Description

Adds HTTP 206 Partial Content support to the Neutralino internal server so media files (e.g., <video>) can be streamed and seeked correctly when served via Neutralino.server.mount().

## Changes proposed

- Detect and parse Range headers in neuserver::handleHTTP.
- Serve only the requested byte range and respond with 206 Partial Content and proper headers.
- Preserve existing behavior for non-range requests.

## How to test it

- Create a Neutralino app with a <video> element served from a mounted path in the resource folder  
- Video is being displayed in partial form

## Next steps

None.

## Deploy notes

None.

#1377


server/neuserver.cpp
```diff
@@ -201,12 +202,48 @@ void handleHTTP(websocketpp::connection_hdl handler) {
...
+                buffer.resize((size_t)length);
+                file.read(&buffer[0],length);
+                std::streamsize byteRead = file.gcount();
+                buffer.resize((size_t)byteRead);
+                con->set_status(websocketpp::http::status_code::partial_content);
+                con->set_body(buffer);
+                con->replace_header("Accept-Ranges","bytes");
+                con->replace_header(
+                    "Content-Range",
+                    "bytes " + std::to_string(start) + "-" + std::to_string(start + byteRead - 1) + "/" + std::to_string(fileSize)
```
shalithasuranga: Please avoid using `std::`

shalithasuranga (changes requested):
Thanks so much for sending this @prasu-baran. Please check my feedback and let me know if you have questions :)

SagarJadhav30 (commented):
best

shalithasuranga (approved):
Looks great. Thanks so much :)

PR #1525, opened 2026-01-19
Fix crash when using main menu with tray on macOS

### LINKED ISSUE #1473: Tray feature causes app crash when Native MacOS main menuitem clicked
Not sure if the Tray feature and Native main menu feature are mutually exclusive but I find that using the main menu will cause an app crash if the tray has started.

If the setTray start is delayed then the main menu will work OK until such time as the tray starts whereupon the app will crash. The crash will occur even if the click handlers are disabled but the tray is still present.
When started the tray will work OK.

No errors left in neutralino.log as a result of crash.

Minimal code -

Main Menu feature -  (note: needs [] to wrap mainMenu - uses .window lib)

    async function setMainMenu() {
    
        let mainMenu = [{
            id: "file", text: "File",
            menuItems: [
                {id: "about", text: "About"}
            ]
        }];

        console.log(Date.now() + ' - setMainMenu called');
        await Neutralino.window.setMainMenu(mainMenu)

    }

Tray feature - (note: does not need [] to wrap tray - uses .os lib)

    async function setTray() {

        let tray = {
            icon: "/resources/trayIcon.png",
            menuItems: [
                {id: "tabout", text: "About"}
            ]
        };
        console.log(Date.now() + ' - setTray called');
        await Neutralino.os.setTray(tray);
    }

Listeners -  (issue will occur even if these are disabled/commented out

//   Neutralino.events.on("mainMenuItemClicked", onMainMenuItemClicked);
//    Neutralino.events.on("trayMenuItemClicked", onTrayMenuItemClicked);

    setMainMenu();
    setTimeout(() => {
        setTray();       (main menu OK until this starts)
    }, 10000);

**Expected behavior**
Main menu items should be clickable without crashing app when tray is present.

**Specifications**  
--- Global ---
neu CLI: v11.6.0 (latest)

--- Project: Hippo (js.neutralino.hippo) ---
Neutralinojs binaries: v6.3.0 (latest)
Neutralinojs client: v6.3.0 (latest)
MacOS 12.7.6 Monterey MBP M1

**Additional context**
Used different id's to avoid potential conflict (eg: "about" - mainmenu & tabout - tray menu). Note: problem is not specific to the about item it will occur with any specified item (eg: licence, readme, etc).

yug105:
Fixes a crash that occurs on macOS when both the main menu (`window.setMainMenu`) and tray (`os.setTray`) features are used           
   together. Clicking any main menu item after the tray has been initialized causes the application to crash with SIGBUS. 
  
Fixes #1473              
                                                                                                                                         
   ## Problem                                                                                                                            
                                                                                                                                         
   On macOS, `NSApplication` can only have **one delegate** at a time. The issue was:                                                    
                                                                                                                                         
   1. When the window initializes, it creates an `AppDelegate` and sets it as the `NSApplication` delegate. This delegate has a          
   `menuCallback:` method that handles main menu item clicks and expects `WindowMenuItem*` data.                                         
                                                                                                                                         
   2. When `os.setTray()` is called, the tray library creates a `Tray` class and sets it as the `NSApplication` delegate,                
   **replacing** the `AppDelegate`.                                                                                                      
                                                                                                                                         
   3. When the user clicks a main menu item, macOS sends `menuCallback:` to the current delegate (now `Tray`). The Tray's                
   `menuCallback:` expects `tray_menu*` data but receives `WindowMenuItem*`, causing a memory access violation (SIGBUS crash).           
                                                                                                                                         
   ## Solution                                                                                                                           
                                                                                                                                         
   Instead of setting the tray as the `NSApplication` delegate, this fix:                                                                
                                                                                                                                         
   - Stores `trayDelegate` in a static variable so it persists                                                                           
   - Sets each tray menu item's **target** explicitly to `trayDelegate` using `setTarget:`                                               
   - Removes the `setDelegate:` call that was overwriting `AppDelegate`                                                                  
   - Adds proper cleanup in `tray_exit()` to release resources                                                                           
                                                                                                                                         
   This way:                                                                                                                             
   - Main menu clicks → go through responder chain → `AppDelegate.menuCallback:` ✓                                                       
   - Tray menu clicks → go directly to target → `Tray.menuCallback:` ✓                                                                   
                                                                                                                                         
   Both menus now work independently without conflicts.                                                                                  
                                                                                                                                         
   ## Testing                                                                                                                            
                                                                                                                                         
   Tested on macOS with the following scenario:                                                                                          
   1. Set main menu with `Neutralino.window.setMainMenu()`                                                                               
   2. Set tray with `Neutralino.os.setTray()`                                                                                            
   3. Click main menu items → No crash, events fire correctly                                                                            
   4. Click tray menu items → No crash, events fire correctly                                                                            
   5. Quit application → Clean exit, no crash


shalithasuranga (approved):
Thanks so much :)

Issue #1203, opened 2024-02-19
WebSocket Server with binary mode support


hschneider:
**Is your feature request related to a problem? Please describe.**
Some situations afford to send and received binary WebSocket messages on an Extension's side. 
E.g. when the WebSocket client implementation does not support text-messages.

**Describe the solution you'd like**
The server should detect incoming text- or binary messages by their opcode.
Then it switches to mode_bin or mode_text for the following conversation.
So if the incoming message is binary, all sent messages to this particular extension will be binary, too.
Other extensions in the same app should still run in their own mode.

**Describe alternatives you've considered**
No alternatives. If a client-implementation  is limited to binary messages, it cannot be used.

**Additional context**
You can put this on low-priority - it's only a nice-to-have, not a show-stopper :-)

hschneider:
Looks like this should improve performance, too:
https://github.com/neutralinojs/neutralinojs/issues/1122

PR #1528, opened 2026-01-20
websocket with binary

prasu-baran:
## Description

This change adds automatic detection of text vs. binary WebSocket frames for extensions from which  server infers the mode from the first incoming message and keeps it per connection, enabling binary-only clients without breaking existing extensions.
## Changes proposed

- Track a per-connection WebSocket mode (text or binary) for extensions.
- Detect the mode from the first incoming frame and use it for all outgoing messages to that extension.

## How to test it

- Create an extension that sends its first message as a binary WebSocket frame.
- Verify that all responses sent back to that extension are also binary, while other extensions still receive text frames.

## Next steps

None.

## Deploy notes

This change is fully backward-compatible with existing text-based extensions.

#1203


shalithasuranga (approved):
Thanks so much :)

PR #1532, opened 2026-01-24
fix: Prevent mount path prefix collision 

### LINKED ISSUE #1468: Mounting path 404 when another similar prefix is mounted
If two paths are mounted like the below, it won't be able to find the files mounted under `testing` - it will 404. In fact, Neutralino seems to crash quite a lot in that scenario.

```js
Neutralino.server.mount('/test', '/var/test');
Neutralino.server.mount('/testing', '/var/testing');
```

I believe this is due to this line of code, where it should be adding a forward slash when checking the path, to properly set a path delimiter so it doesn't accidentally match other mounted paths with same prefix:

https://github.com/neutralinojs/neutralinojs/blob/3048628183a2950e0c618524afb2f55ccde09455/server/router.cpp#L353

Maybe it should be this, with additional tweaks to other code?

```c++
if(pathname.find(mountedPath + '/') == 0) {
```

malladinagarjuna1:
##  Problem
Fixes #1468

When two server mount paths share similar prefixes, requests to the longer path incorrectly match the shorter one.

### Example

```js
Neutralino.server.mount('/test', '/var/test');
Neutralino.server.mount('/testing', '/var/testing');
```

A request to:

```
/testing/file.txt
```

Incorrectly matches:

```
/test
```

This results in:

* Serving from the wrong directory
* 404 errors
* Potential crashes

---

##  Root Cause

The existing condition:

```cpp
pathname.find(mountedPath) == 0
```

Matches **any path that starts with the mount path**, without validating proper path boundaries.

So:

```
/testing
```

incorrectly matches:

```
/test
```

---

## Solution

Updated the mount path matching logic in `router.cpp` to include a **path delimiter check**.

Now, a match is only valid if it is:

* An **exact match**

  ```
  /test == /test
  ```
* Or followed by a **path delimiter**

  ```
  /test/file.txt
  ```

This prevents false matches like:

```
/testing
```

matching:

```
/test
```

---

---

##  Testing

Added a regression test in `server.spec.js`:

* Mounts both `/test` and `/testing`
* Verifies each path serves files from the correct directory
* Confirms no incorrect cross-matching

All existing tests pass successfully.

---

##  Files Changed

* **router.cpp** — Core fix for mount path matching

* **server.spec.js** — Added regression test for similar mount paths

---

### DIFF
--- a/spec/server.spec.js
+++ b/spec/server.spec.js
@@ -71,6 +71,33 @@ describe('server.spec: server namespace tests', () => {
             assert.ok(typeof output === 'object', 'Expected output is an object');
             assert.ok(output.fetch1 === 200, 'The file request to a mounted directory succeeds');
         });
+
+        it('handles similar prefix mount paths correctly', async () => {
+            runner.run(`
+                const response = {};
+                const testPath = NL_PATH + '/.tmp/test';
+                const testingPath = NL_PATH + '/.tmp/testing';
+                
+                await Neutralino.filesystem.createDirectory(testPath);
+                await Neutralino.filesystem.createDirectory(testingPath);
+                await Neutralino.filesystem.writeFile(testPath + '/file.txt', 'From /test');
+                await Neutralino.filesystem.writeFile(testingPath + '/file.txt', 'From /testing');
+                
+                await Neutralino.server.mount('/test', testPath);
+                await Neutralino.server.mount('/testing', testingPath);
+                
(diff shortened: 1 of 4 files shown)


spec/server.spec.js
prasu-baran: here hardcoded /testing and /test is being used it is not expected to go like this
You can think of making it universal for other file paths too to be accesed via this

PR #1533, opened 2026-01-26
Fix macOS crash on exit

### LINKED ISSUE #1469: Neutralino.app.exit() throws NSException on Mac
**Describe the bug**
The command Neutralino.app.exit() throws NSException on Mac.
Even though the window closes, the error leaves a long stack trace in the console and/or opens Mac's crash report window every time the application gets closed.

**To Reproduce**
1. Initiate a simple neutralinojs project (v6.3.0) runnable in window mode,
2. Set `modes.window.exitProcessOnClose` to `false` in `neutralino.config.json`,
3. Add the following snippet to main.ts (or an equivalent file):
```
async function onWindowClose() {
    console.log('Saving data and closing the window...')
    await Neutralino.app.exit()
}

// Initialize Neutralino
Neutralino.init()

// Register event listeners
Neutralino.events.on('windowClose', onWindowClose)
```
4. Close the application by clicking on X button and observe the error being thrown.

**Expected behavior**
The tool runs onWindowClose function and exits gracefully.

Image

Output in the console (without full stack trace):

> 2025-10-02 15:56:37.608 neutralino-mac_arm64[23540:187414] This application is modifying the autolayout engine from a background thread after the engine was accessed from the main thread. This can lead to engine corruption and weird crashes.
>  Stack:(**[...]**)
> *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Modifications to the layout engine must not be performed from a background thread after it has been accessed from the main thread.'
> *** First throw call stack:
> (**[...]**)
> libc++abi: terminating due to uncaught exception of type NSException
> neu: INFO neutralino-mac_arm64 was stopped with success code 0

Full stack report attached [here](https://github.com/user-attachments/files/22662480/full_stack_trace.txt).

**Specifications**  
 - OS: macOS Tahoe v26.0.1
 - Neutralinojs version: v6.3.0
 - Neutralinojs client library version: v6.3.0
 - Neutralinojs CLI version: v11.6.0

**Additional context**
I was using version 5.5.0 and the exit command was working correctly After switching to 6.3.0 the problem showed up. The error message suggests that there is a threading issue in MacOS.

I was able to find a workaround. I use `await Neutralino.app.killProcess()` instead `exit()`. Although this approach is not recommended according to the documentation.

Thanks in advance for looking into it!

Namiiii-sys:
Fixes #1469 

## Description

This PR addresses a crash occurring on macOS when Neutralino.app.exit() is called.

The Root Cause: Threading Violation: app.exit() is triggered from the background WebSocket thread, but AppKit requires that UI operations specifically saving window state and terminating the window occur exclusively on the Main Thread.

## Changes proposed

 - api/window/window.cpp: Wrapped the _close logic in a dispatch_sync block targeting the Main Queue on macOS. This ensures UI operations happen on the correct thread.
 

## How to test it

  - Build the project on macOS (ARM64/x64) using cmake .
  - Run the generated binary.
  - Open the Developer Tools (Inspector).
  - Execute Neutralino.app.exit() in the console.
  - Expected Result: The application closes instantly and cleanly, with no stack trace or error output in the terminal.

Note on Iteration: In my previous commit, I attempted to resolve the crash by removing the close() call entirely to avoid a suspected double-free. After further testing and feedback, I've refined the fix to focus strictly on the threading violation. This current version keeps all original cleanup logic intact but ensures it executes on the Main Thread, preventing the crash while avoiding any risk of "zombie" processes.

### DIFF
--- a/api/window/window.cpp
+++ b/api/window/window.cpp
@@ -693,6 +693,16 @@ bool __createWindow() {
 
 void _close(int exitCode) {
     if(nativeWindow) {
+        #if defined(__APPLE__)
+        dispatch_sync(dispatch_get_main_queue(), ^{
+            if(windowProps.useSavedState) {
+                __saveWindowProps();
+            }
+            nativeWindow->terminate(exitCode);
+            delete nativeWindow;
+            nativeWindow = nullptr;
+        });
+        #else   
         if(windowProps.useSavedState) {
             __saveWindowProps();
         }
@@ -701,6 +711,8 @@ void _close(int exitCode) {
         FreeConsole();
         #endif
         delete nativeWindow;
+        nativeWindow = nullptr;
+        #endif
     }
 }
 
--- a/api/window/window.cpp
+++ b/api/window/window.cpp
@@ -12,6 +12,7 @@
 
 #elif defined(__APPLE__)
 #include <objc/objc-runtime.h>
+#include <dispatch/dispatch.h>
 #include <CoreFoundation/Corefoundation.h>
 #include <CoreGraphics/CGDisplayConfiguration.h>
 #include <CoreGraphics/CGWindow.h>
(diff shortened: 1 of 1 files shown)


shalithasuranga (approved):
Thanks so much :tada: I've updated the suggested code a bit

PR #1534, opened 2026-01-26
 Add Ctrl+W/Cmd+W keyboard shortcut to close application window 

lahari936:
- Implements standard close window shortcut for all platforms
- Uses Ctrl+W on Windows/Linux and Cmd+W on macOS
- Prevents default browser behavior
- Integrates with Neutralino.app.exit() API
- Platform detection via navigator.platform

## Changes proposed

 - Change 1
 - Change 2
 - etc..

## How to test it

 - Run specs/tests

## Next steps

None.

## Deploy notes

None.

### DIFF
--- a/bin/resources/js/main.js
+++ b/bin/resources/js/main.js
@@ -30,4 +30,15 @@ Neutralino.events.on("eventFromExtension", (evt) => {
     console.log(`INFO: Test extension said: ${evt.detail}`);
 });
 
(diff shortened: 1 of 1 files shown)


shalithasuranga:
Hello, thanks so much for sending this pull request. I had to close this since we don't update the simple sample app with features -- it's only there for testing and development purposes. I think the above-mentioned feature belongs to the app -- not the framework.

PR #1543, opened 2026-02-08
Rewrite router.cpp/getAsset to work with files without extensions

### LINKED ISSUE #1542: mount semantics are incorrect when resolving a file with no extension
**Describe the bug**
When using `Neutralino.server.mount`, files with no extension fail to resolve with a 404 error.  This is because when no `.` is present in the filename, the `getAsset` router code assumes it is resolving a folder, and appends `/index.html`.

**To Reproduce**
1. Use Neutralino.server.mount to mount a filesystem directory that contains a file with no `.` in its filename.
2. Attempt to load the file with `fetch`, url in css, or an HTML attribute.
3. Observe error in console reporting a 404.

browser:
```
Uncaught (in promise) Error: Failed to load image /project/cards/Beast, undefined
```
stdout:
```
[app] ERROR 2026-02-07 23:14:19,969 NE_RS_UNBLDRE: Unable to load application resource file /resources/project/cards/Beast/index.html api/debug/debug.cpp:20 mator@unknown-host
```

**Expected behavior**
Files should resolve even if they do not have an extension.  Filesystems do not offer any guarantee about `.` -> file.

**Specifications**  
 - OS: Manjaro
 - Neutralinojs version: v6.5.0
 - Neutralinojs client library version: v6.5.0
 - Neutralinojs CLI version: v11.7.0

**Additional context**
If support for serving extensionless files is added, it will also be prudent to resolve MIME-types based on file magic, rather than extension.

matortheeternal:
## Description
This PR rewrites `router::getAsset` to not append  `/index.html` when paths do not contain `.`, and instead only append it when the provided path resolves to a directory.

## Changes proposed
- Resolve +`/index.html` from mounts when the path resolved on the filesystem is a directory.
- Resolve +`/index.html` from resources when the path resolved in the embedded assets, assets bundle, or assets folder is a directory.
- Rewrite `getAsset` to not be recursive - SPA resolution now invokes the `resolveFileFromMounts` and `resolveResourceFile` helper functions instead of relying on recursive behaviors which do not accurately map to this problem.
- Use libmagic on apple and linux to resolve MIME type from the file buffer in memory instead of relying on the file extension.
  - It may be prudent to use libmagic on windows as well for consistency across platforms.

## How to test it
This PR resolves #1542.

 - Run specs/tests
 - Test asset resolution for mounts, bundled resources, embedded resources, folder resources, and SPA mode, including:
   - files with extensions
   - files without extensions
   - folders
   - folders with extension-like names

## Next steps
I am looking for discussion and buy-in on these changes. Given the fact that this proposal is adding a dependency, I understand it may take some time and consideration for acceptance.

I am open to discussion on alternative implementations which are more in line with existing priorities within the Neutralino codebase and dependency ecosystem.

## Deploy notes
None.


matortheeternal:
SPA (`--single-page-serve`) behavior remains untested in the codebase, and this PR does not attempt to define or extend SPA semantics beyond existing behavior.

Given the lack of test coverage, I have not added new SPA-specific tests here. I am open to guidance on adding coverage if deemed necessary and desirable by the project maintainers.

shalithasuranga:
Thanks so much for sending this pull request. We would like to keep the current static server implementation since we don't need a fully-featured static server -- just need to serve some common web app files. Thanks :tada:

matortheeternal:
@shalithasuranga I don't understand the reason.  Is an alternative implementation going to be used to resolve #1542 ?

Issue #1548, opened 2026-02-11
macOS: window.focus() doesn't bring window to front when another app is active


uneco:
**Describe the bug**

`Neutralino.window.focus()` on macOS doesn't bring the window to front when the active app's window is in front.

**To Reproduce**

1. Put another app's window in front of the Neutralino app window
2. Call `window.focus()`
3. If the front window has focus: Neutralino window stays behind
4. If the front window doesn't have focus: Neutralino window comes to front

**Expected behavior**

The documentation says:
- `window.show()`: "Shows the native window."
- `window.focus()`: "Focuses the native window."

The difference between these two is unclear. However, looking at other platforms' implementations, Linux uses `gtk_window_present()` and Windows uses `SetForegroundWindow()`, both of which activate the window and bring it to front. The macOS implementation only calls `orderFront:`, which cannot bring a window above the active application's windows.

There is no Neutralino API on macOS that can activate a window when another app is active.

**Specifications**
- OS: macOS
- Neutralinojs version: v6.5.0

**Additional context**

Current macOS implementation (`api/window/window.cpp`):
```cpp
((void (*)(id, SEL, id))objc_msgSend)((id) windowHandle,
        "orderFront:"_sel, NULL);
```

Possible fix using `activate` + `makeKeyAndOrderFront:`:
```cpp
((void (*)(id, SEL))objc_msgSend)(
    ((id(*)(id, SEL))objc_msgSend)("NSApplication"_cls, "sharedApplication"_sel),
    "activate"_sel);
((void (*)(id, SEL, id))objc_msgSend)((id) windowHandle,
        "makeKeyAndOrderFront:"_sel, NULL);
```

PR #1549, opened 2026-02-11
fixed(#1548): window.focus() brings window to front on macOS

uneco:
## Description

Fix `Neutralino.window.focus()` on macOS to bring the window to front when another app is active.

Fixes https://github.com/neutralinojs/neutralinojs/issues/1548

## Changes proposed

- Call `[NSApplication sharedApplication].activate` to activate the app
- Use `makeKeyAndOrderFront:` instead of `orderFront:` to make the window key window and bring it to front

### Why this fix works

The previous implementation only called `orderFront:`, which did not bring the window to front when another app was active. The fix calls `activate` first, then `makeKeyAndOrderFront:`. This matches `gtk_window_present()` on Linux and `SetForegroundWindow()` on Windows.

## How to test it

1. Launch the Neutralino app
2. Put another app's window in front of the Neutralino app window
3. Call `Neutralino.window.focus()`
4. Verify the Neutralino window comes to the front

### DIFF
--- a/api/window/window.cpp
+++ b/api/window/window.cpp
@@ -1432,8 +1432,11 @@ json focus(const json &input) {
     #if defined(__linux__) || defined(__FreeBSD__)
     gtk_window_present(GTK_WINDOW(windowHandle));
     #elif defined(__APPLE__)
+    ((void (*)(id, SEL))objc_msgSend)(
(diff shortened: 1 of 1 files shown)


shalithasuranga (commented):
Thanks so much @uneco :tada:

PR #1551, opened 2026-02-12
Add windowActivate event for macOS dock icon click

uneco:
## Description

macOS users expect clicking the dock icon to bring a hidden window back. Neutralino has no event for this.

This PR adds a `windowActivate` event that fires when the user clicks the dock icon, via the `applicationShouldHandleReopen:hasVisibleWindows:` delegate method.

## Why the tray library is also changed

This was supposed to be a one-line delegate method addition to `AppDelegate`. But `lib/tray/tray.h` calls `[NSApp setDelegate: trayDelegate]` in `tray_init()`, which replaces the `AppDelegate` that `webview.h` already set. So `applicationShouldHandleReopen:hasVisibleWindows:` never gets called.

`AppDelegate` is the correct NSApp delegate, not `Tray`. This PR removes the `setDelegate:` call from `tray_init()` and uses explicit `setTarget:` on each `NSMenuItem` to route tray menu callbacks directly to `trayDelegate`.

## Changes

- Add `WEBVIEW_WINDOW_ACTIVATE` constant in `lib/webview/webview.h`
- Add `NSApplicationDelegate` protocol to `AppDelegate`
- Register `applicationShouldHandleReopen:hasVisibleWindows:` on `AppDelegate`
- Dispatch `windowActivate` from `windowStateChange` in `api/window/window.cpp`
- Remove `[NSApp setDelegate: trayDelegate]` from `lib/tray/tray.h`
- Use `setTarget:` on tray menu items instead of relying on the responder chain

## How to test

1. Build and run a Neutralino.js app on macOS
2. Call `Neutralino.window.hide()` or switch to another app
3. Click the dock icon
4. Confirm `windowActivate` fires via `Neutralino.events.on('windowActivate', ...)`
5. Verify tray menu items still work

## Deploy notes

Merge together with neutralinojs/neutralino.js#190 (TypeScript type definition).

### DIFF
--- a/lib/tray/tray.h
+++ b/lib/tray/tray.h
@@ -176,6 +176,7 @@ static id pool;
 static id statusBar;
 static id statusItem;
 static id statusBarButton;
+static id trayDelegate;
 
 static id _tray_menu(struct tray_menu *m) {
   id menu = ((id (*)(id, SEL))objc_msgSend)(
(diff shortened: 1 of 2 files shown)


shalithasuranga:
Thanks so much for sending this @uneco :tada:

PR #1552, opened 2026-02-21
fix: use atomic counter for openFile IDs to prevent reuse after close

Aman-Cool:
**Description**

When you open and close files using `filesystem.openFile`, file IDs are generated using `openedFiles.size()`. This works until any file is closed — the map shrinks, the next open reuses an existing ID, and the old `ifstream*` is silently overwritten. Reads from the victim ID now return data from the wrong file with no error.

This PR replaces the `size()` call with an `atomic<int>` counter, the same pattern already used for process IDs in `os.cpp`.

---

**Why the counter approach is needed**

`openedFiles.size()` only produces unique IDs if entries are never removed. The moment `updateOpenedFile` closes a file and erases the map entry, the next `openFile` call computes an ID that may already be occupied. `std::map::operator[]` on an existing key silently replaces the value — no error, no warning, wrong file handle from that point on. The leaked `ifstream*` also holds the OS file descriptor open until process exit, locking the file on Windows.

---

**Changes**

- Add `atomic<int> nextVirtualFileId(0)` alongside the existing `openedFiles` map in `api/fs/fs.cpp`
- Replace `openedFiles.size()` with `nextVirtualFileId++` in `fs::openFile()`

---

**How to test**

1. Open three files via `Neutralino.filesystem.openFile`
2. Close the first one via `updateOpenedFile` with `event: 'close'`
3. Open a fourth file — note the returned ID
4. Read from the third file's ID
5. Confirm data comes from the correct file, not the newly opened one
6. Verify no file descriptor leak across repeated open/close cycles

---

**Deploy notes**

Companion type definition update may be needed in `neutralinojs/neutralino.js` if `openFile` return types are documented there.

### DIFF
--- a/api/fs/fs.cpp
+++ b/api/fs/fs.cpp
@@ -197,12 +197,12 @@ bool writeFile(const fs::FileWriterOptions &fileWriterOptions) {
 }
 
 int openFile(const string &filename) {
-    int virtualFileId = nextFileId++;
     ifstream *reader = new ifstream(CONVSTR(filename), ios::binary);
     if(!reader->is_open()) {
         delete reader;
         return -1;
     }
+    int virtualFileId = nextFileId.fetch_add(1);
     lock_guard<mutex> guard(openedFilesLock);
     openedFiles[virtualFileId] = reader;
     return virtualFileId;


prasu-baran:
@Aman-Cool  nice implementation but this increment which you added will in long term cause problem for us as even for no  file opened it is still incrementing so I would suggest to perform this increment at later after return -1 this would be more appropriate

api/fs/fs.cpp
```diff
@@ -195,7 +197,7 @@ bool writeFile(const fs::FileWriterOptions &fileWriterOptions) {
 }
 
 int openFile(const string &filename) {
-    int virtualFileId = openedFiles.size();
+    int virtualFileId = nextFileId++;
```
SharonIV0x86:
```suggestion
    int virtualFileId = nextFileId.fetch_add(1);
```

Thoughts?

api/fs/fs.cpp
```diff
@@ -195,7 +197,7 @@ bool writeFile(const fs::FileWriterOptions &fileWriterOptions) {
 }
 
 int openFile(const string &filename) {
```
SharonIV0x86: It seems like you are incrementing the counter a bit early.
There is a possibility that the file opening operation fails at this point 

```cpp
if(!reader->is_open()) {
        delete reader;
        return -1;
    }
```
So you "consume" a file ID even though no file was successfully opened.
It would be better if you increment and assign after the if condition.

api/fs/fs.cpp
Aman-Cool: Makes sense! Though `nextFileId++` on `std::atomic<int>` is equivalent, the `++` operator is overloaded to do an atomic fetch_add internally. Happy to apply it for explicitness if you prefer, but I'll also be moving the increment to after the `is_open()`.

PR #1554, opened 2026-02-21
refactor: Refactored storage tests.

SharonIV0x86:
## Description
The current storage tests in ``/spec/storage.spec.js`` are coupled together. Tests share a common ``.storage``(sharing a common dir is fine) directory without cleanup. Some tests depend on keys created by previous tests. Reordering tests or running them individually may cause failures. In future if we run the test cases in parallel it can cause failures due to test dependency. Or if the order of execution of tests is change it can also cause failures.

This matters because, test results should not depend on execution order. Future contributors adding tests could unintentionally break existing ones

## Changes proposed
- Added beforeEach hook to clear storage before each test.
- Removed order-dependent assertions.
- Ensured each test verifies only what it creates.
- Made storage.getKeys assertions deterministic.

## How to test it
go to the /spec folder.
run the following commands in order.
- npm install
- npm run test storage

## Deploy notes
None.

I have tried to keep the tests behavior similar to the ones before.

### DIFF
--- a/spec/storage.spec.js
+++ b/spec/storage.spec.js
@@ -1,46 +1,51 @@
 const assert = require('assert');
-
 const runner = require('./runner');
 
 describe('storage.spec: storage namespace tests', () => {
 
+    beforeEach(() => {
+        runner.run(`
+            await Neutralino.storage.clear();
+            await __close('done');
+        `);
+        assert.strictEqual(runner.getOutput(), 'done');
+    });
+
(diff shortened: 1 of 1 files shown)


shalithasuranga:
We would like to keep `assert.ok` to be consistent with other parts of the test suite. We can improve the whote test suite later if needed. Thanks for sending this :tada:

PR #1556, opened 2026-02-21
feat(webview): open window.open / NewWindowRequested URLs in system browser (Windows)

### LINKED ISSUE #1555: Handle window.open / NewWindowRequested by opening URL in system browser
**Description** 
When the app runs in window mode, calls to window.open(url) or links with target="_blank" are handled by the embedded WebView’s default behavior (e.g. in-place navigation or an empty/popup window). There is no way to have these open in the user’s default system browser instead.|

**Solution which can be implemented** 
Handle the WebView’s “new window requested” event and open the requested URL in the system default browser (same behavior as Neutralino.os.open(url)), and mark the event as handled so the WebView does not open its own window.
**Windows (WebView2):** Implement ICoreWebView2NewWindowRequestedEventHandler. In the handler: get URI from the event → ShellExecuteW(nullptr, L"open", uri, ...) → put_Handled(TRUE).
**Linux (GTK/WebKit):** Connect to the WebKit “create” / new-window signal → get URL → xdg-open (or equivalent) → indicate handled so WebKit doesn’t create a new view.
**macOS (WKWebView):** Implement the WKUIDelegate method for new window requests → get URL → open with system (e.g. open command) → cancel default new window.

**Alternatives which can be considered to be used**
Using Neutralino.os.open(url) instead of window.open(url): works but requires changing all window.open and target="_blank" usage in the app and doesn’t help generic links.
Leaving current behavior: window.open and target="_blank" keep using WebView default behavior, which is often not desirable for external links.

**Benefits**
Same frontend code works in PWA and desktop: window.open() and target="_blank" links open in the system browser in the desktop build.
No need to replace every window.open / target="_blank" with Neutralino.os.open or custom click handlers.

malladinagarjuna2:
**Description**
fixes #1555 
**Summary**

When the app calls window.open(url) or the user clicks a link with target="_blank", the requested URL is now opened in the system default browser instead of using the WebView’s default behavior (e.g. in-place navigation or empty popup).
Problem
Neutralino did not handle the WebView’s “new window requested” event. As a result:
window.open("https://...") and target="_blank" links did not open in the system browser.
Apps had to use Neutralino.os.open(url) and custom click handlers instead of standard web behavior.

**Solution**
Windows (WebView2): Implemented ICoreWebView2NewWindowRequestedEventHandler. On new-window requests we:
Read the URI from the event.
Open it with ShellExecuteW(..., L"open", uri, ...) (system default browser).
Call put_Handled(TRUE) so the WebView does not open its own window.

**Changes**
lib/webview/webview.h
Included <shellapi.h> and linked Shell32.lib.
webview2_com_handler now implements ICoreWebView2NewWindowRequestedEventHandler.
Registered with webview->add_NewWindowRequested(this, &token).
New Invoke(..., ICoreWebView2NewWindowRequestedEventArgs*) handler that opens the URI in the system browser and marks the event as handled.

**Testing**
Build and run the app in window mode.
In the page, run window.open("https://example.com") or click a link with target="_blank" to an external URL.
The URL should open in the default system browser (e.g. Chrome/Edge), and no extra window should appear in the app.
Platform

**Windows:** Implemented (WebView2).
**Linux (GTK/WebKit) and macOS (Cocoa):** Not included in this PR; can be added in a follow-up.

**References**
**MDN:** Window.open()
**WebView2:** NewWindowRequested

### DIFF
--- a/lib/webview/webview.h
+++ b/lib/webview/webview.h
@@ -745,12 +745,14 @@ using browser_engine = cocoa_wkwebview_engine;
 
 #define WIN32_LEAN_AND_MEAN
 #include <Shlwapi.h>
+#include <shellapi.h>
 #include <codecvt>
 #include <stdlib.h>
 #include <windows.h>
 
 #pragma comment(lib, "user32.lib")
 #pragma comment(lib, "Shlwapi.lib")
+#pragma comment(lib, "Shell32.lib")
 
 // Edge/Chromium headers and libs
 #include "webview2.h"
@@ -888,7 +890,8 @@ private:
       : public ICoreWebView2CreateCoreWebView2EnvironmentCompletedHandler,
         public ICoreWebView2CreateCoreWebView2ControllerCompletedHandler,
         public ICoreWebView2WebMessageReceivedEventHandler,
-        public ICoreWebView2PermissionRequestedEventHandler {
+        public ICoreWebView2PermissionRequestedEventHandler,
+        public ICoreWebView2NewWindowRequestedEventHandler {
     using webview2_com_handler_cb_t =
         std::function<void(ICoreWebView2Controller *)>;
 
@@ -915,6 +918,7 @@ private:
       controller->get_CoreWebView2(&webview);
       webview->add_WebMessageReceived(this, &token);
       webview->add_PermissionRequested(this, &token);
+      webview->add_NewWindowRequested(this, &token);
 
       m_cb(controller);
       return S_OK;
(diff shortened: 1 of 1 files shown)


malladinagarjuna2:
ok will implement that too but i don't have macos to test  it

malladinagarjuna2:
@shalithasuranga  sir could you please review my pr

shalithasuranga:
We need to discuss this behavior by comparing it with other frameworks. Closing until we finish the discussions. Thanks so much :tada:

PR #1557, opened 2026-02-21
fix: Fix CI failure on darwin due to ``window.snapshot``.

SharonIV0x86:
## Description
I noticed that the CI tests are failing on darwin due to ``window.snapshot``. ``window.snapshot`` being the topmost test is failing and all the consecutive ``window`` tests are getting skipped.

## Changes proposed
This PR updates the macOS implementation of ``window.snapshot`` to use the modern ``ScreenCaptureKit`` workflow in a way that is compatible with the macOS 15 SDK. The previous implementation attempted to construct an ``SCWindow`` instance directly using a selector that is no longer available at runtime on recent macOS versions, which caused crashes in CI when building against newer SDKs. 

The new implementation retrieves the list of shareable windows via ``SCShareableContent``, selects the current window by its ``windowNumber``, and performs the capture using ``SCScreenshotManager``. The asynchronous capture API is synchronized using a semaphore so that the existing synchronous function contract is preserved. 

This resolves the macOS CI failure while keeping the implementation aligned with Apple’s current ``ScreenCaptureKit`` usage model.

## How to test it
Run the CI.

 - Run specs/tests

## Next steps
Check if the snapshot functionality works as intended. And fix the code formatting.

None.

## Deploy notes

None.

### DIFF
--- a/api/window/window.cpp
+++ b/api/window/window.cpp
@@ -1093,33 +1093,33 @@ bool snapshot(const string &filename) {
     GdkPixbuf *screenshot = gdk_pixbuf_get_from_window(window, x, y, width, height);
     return gdk_pixbuf_save(screenshot, filename.c_str(), "png", nullptr, nullptr);
 
-#elif defined(__APPLE__)
+    #elif defined(__APPLE__)
 
-if (@available(macOS 12.3, *)) {
+    if (@available(macOS 12.3, *)) {
 
-    long winId =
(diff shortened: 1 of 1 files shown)


prasu-baran:
no issues just referring if want help can see the codebase

shalithasuranga (approved):
Thanks so much :)

PR #1560, opened 2026-02-24
fix: memory leaks in the tray management system

NipunSGeeTH:
## Description
This pull request fixes persistent memory leaks in the tray management system. Previously, tray icon resources (like HICON on Windows and NSImage on macOS) were allocated every time the tray was updated but were never released, leading to increasing memory consumption.

## Changes proposed

   Added platform-specific cleanup logic in setTray
 to release the previous icon before allocating a new one.
Implemented proper resource disposal in cleanupTray to ensure all icon memory is freed when the application closes or the tray is removed.

## How to test it
- Initialize the tray and update the icon multiple times via the API.
- Monitor the application's memory usage to ensure it remains stable instead of growing with each update.
- Close the application and verify that tray resources are correctly released.

 - Run specs/tests

## Next steps

None.

## Deploy notes

None.

### DIFF
--- a/api/os/os.cpp
+++ b/api/os/os.cpp
@@ -88,6 +88,20 @@ bool isTrayInitialized() {
 void cleanupTray() {
     if(os::isTrayInitialized()) {
         tray_exit();
+#if defined(_WIN32)
+        if (tray.icon) {
+            DestroyIcon(tray.icon);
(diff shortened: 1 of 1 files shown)


shalithasuranga (approved):
Thanks so much for your contribution :tada:

PR #1562, opened 2026-02-24
Fix race condition in updateSpawnedProcess due to missing mutex lock

### LINKED ISSUE #1474: os.updateSpawnedProcess stdInEnd action exits spawned process - part 2
Not so sure bug #961 has been fixed (I don't have permission to reopen it).

Revisiting this with v6.3.0 and I still get the issue.

Scenario: 

sqlite3 executable is started via os.spawnProcess and the sqlite3 process can be seen in the Mac Process Monitor

Request .help - 

    await Neutralino.os.updateSpawnedProcess(sqliteId, 'stdIn', '.help')
            
Of course to obtain the data returned by .help via stdout I need to do -

    await Neutralino.os.updateSpawnedProcess(sqliteId, 'stdInEnd');

The data is retrieved but the stdInEnd appears to trigger an exit request and the sqlite3 process is killed.

Looking at the docs the possible actions are -

    stdIn: Sends data to the process via the standard input stream.
    stdInEnd: Closes the standard input stream file descriptor.
    exit: Terminates the process.

I expected stdInEnd to simply close/reset the stdin so the stdin is then available for further input to the existing spawned process. 
If the stdInEnd is going to kill the process what is the point of the exit action which I expect should kill the process if used.

If the stdInEnd is intended to reset the stdIn then I would assume that the next os.updateSpawnedProcess() will open the stdIn for further use.

Looking at some of the source I feel that stdInEnd should not kill the process.

Neu source sample -

Image

Tiny-Process-Library sample -

Image

Specifications
--- Global ---
neu CLI: v11.6.0 (latest)

--- Project: NeuSQLite (uk.wanderinghippo.neusqlite) ---
Neutralinojs binaries: v6.3.0 (latest)
Neutralinojs client: v6.3.0 (latest)
MacOS 12.7.6 Monterey MBP M1

nagadevikona20-max:
## Description
`updateSpawnedProcess` was accessing the shared `spawnedProcesses` map and dereferencing
the raw process pointer without holding `spawnedProcessesLock`. This creates a race
condition where the background `processThread` can delete the pointer between the map
lookup and the method call in `updateSpawnedProcess`, leading to undefined behavior.
Every other function that touches `spawnedProcesses` already acquires the lock correctly.

Fixes #1474

## Changes proposed

 - Added `lock_guard<mutex> guard(spawnedProcessesLock)` at the start of `updateSpawnedProcess` to match the locking pattern used consistently across `spawnProcess`, `processThread`, and `getSpawnedProcesses`

## How to test it

 - Spawn a process that reads from stdin like `sqlite3` or `cat`
 - Send input via `stdIn` and then call `stdInEnd` in quick succession
 - Confirm the process handles the stdin close without a crash or unexpected behavior from a corrupted pointer

## Next steps

None.

## Deploy notes

None.

### DIFF
--- a/api/os/os.cpp
+++ b/api/os/os.cpp
@@ -216,6 +216,7 @@ pair<int, int> spawnProcess(string command, const os::ChildProcessOptions &optio
 }
 
 bool updateSpawnedProcess(const os::SpawnedProcessEvent &evt) {
+    lock_guard<mutex> guard(spawnedProcessesLock);
     if(spawnedProcesses.find(evt.id) == spawnedProcesses.end()) {
         return false;
     }


shalithasuranga (approved):
Thanks so much :tada:

PR #1563, opened 2026-02-24
feat(fs): implement access, chmod, and chown

### LINKED ISSUE #1372: A variant of `fs.access` and `fs.chmod` for Neutralino.filesystem API
## Is your feature request related to a problem? Please describe

There are cases where you need to check whether folders or files are writable/readable.

* For example, in my app, I need to create and manage a folder for user data before my app can do anything. The default folder is usually $HOME/ct.js, but:
  * Apparently on Windows there are issues for Active Directory users where their home directory links to a remote SAMBA drive;
  * And Linux is also a zoo that can't be trusted.
  * So I have to check whether a directory is empty and allow users to pick another one. For that, I need some sort of `fs.access` method from Node.js.
* Executables that don't have +x flag seems to be a reoccuring theme in various places nowadays 🤔 

These cases require a way to read folder and file access right, and `fs.chmod`-like method would allow apps built on Neutralino to mend these situations automatically.

## Describe the solution you'd like
An implementation for Node.js' [`fsPromise.access`](https://nodejs.org/api/fs.html#fspromisesaccesspath-mode) and [`fsPromise.chmod`](https://nodejs.org/api/fs.html#fspromiseschmodpath-mode) for Neutralino.js.

### `fs.access`
"R" means "read access", "W" means "write access", "X" means "can run as executable".

Using the existing filesystem API:
* Checks for R for files can be replaced with `filesystem.readBinaryFile`, but this brings an overhead for large files as the whole content of a file is transmitted.
* Checks for R for directories is `filesystem.readDirectory`, same concerns with overhead.
* Checks for W for files are tricky and fails on files that are locked on Windows. _Close to impossible._ You need to somehow re-write the file without any possible dependants on it notice that.
* Checks for W for folders is usually creating an empty temp file. If you can create it — nice, the folder is writable.
* **Checks for X is not possible with the current API.**

You can also [use `stat`](http://askubuntu.com/questions/152001/ddg#152003) with `os.execCommand` on Linux/macOS…

### `fs.chmod`

* No alternatives besides from using CLI `chmod` with `os.execCommand` on Linux/macOS 🤷 

## Additional context

* `fs.chmod` makes sense for Linux and macOS only. It is usually unimplemented on Windows.
* Maybe add `fs.chown` for the sake of completeness for Linux/macOS, too?

SagarJadhav30:
***Description***

This commit adds support for filesystem.access, filesystem.chmod, and filesystem.chown, mirroring the functionality found in the Node.js fs module. These methods let users query permissions and change file access directly via the Neutralinojs C++ backend.

fixes : #1372

***Changes proposed***

Implemented filesystem.access, filesystem.chmod, and filesystem.chown in api/fs
Added the corresponding error mappings (NE_FS_ACCERR, NE_FS_CHMDERR, NE_FS_CHWNERR) to 

errors.cpp
 and 

errors.h
Registered the native functions in 

server/router.cpp

***How to test it***

Build the C++ backend using neu build
Use an application to invoke the new filesystem.access, filesystem.chmod, and filesystem.chown native API methods directly against varying local files and folders to verify functionality and error codes.

***Next steps***

Update the javascript bindings in the neutralino.js client library repository to provide access to these new functions.

***Deploy notes***
None.


shalithasuranga (approved):
Thanks so much :tada:

PR #1564, opened 2026-02-24
fix(windows): restore keyboard focus on window activation (#1491)

### LINKED ISSUE #1491: No keyboard focus when the window is focused. Neutralino requires an initial click inside the app. This makes many apps not viable
**Describe the bug**
On Windows, the keyboard input is ignored when the Neutralino app window is focused. It only starts working after the user clicks on the html document.  
Forcing focus via JS doesn't solve it.
It's easy to reproduce. **Mac OS doesn't have this issue.**
Linked a short YouTube video below.
  
  
**To Reproduce**
1. Create a new project using neu create ...
2. Test the keyboard input by pasting this code at the bottom of main.js
```
document.addEventListener("keyup", (e) => {
  console.log("pressed:", e.key);
  alert("pressed:" + e.key);
});
```
3. First check that it works by clicking inside the app, and pressing a keyboard letter.
4. Click outside the app (like the Windows taskbar or the desktop) so that the app loses focus.
5. Refocus the window by clicking its taskbar icon, or the window title bar (NOT the html document).
6. Press the keyboard keys and notice how they are ignored.
7. Click inside the app's html document.
8. Press the keyboard keys and notice how they now work.
9. Optionally, test the same on Mac OS to compare. There's no problem there.
  
  
**Alternative way to reproduce**
1. Add an input to the html:
`<input type="text" />`
2. Try forcing focus on the input in main.js:
 ```
Neutralino.events.on("windowFocus", () => {
  console.log("window focused");
  forceFocus();
});

function forceFocus() {
  setTimeout(() => {
    //document.body.focus();
    document.querySelector("input").focus();
  }, 500);
}
```
  
  
**Expected behavior**
When the app's window is focused, the keyboard input should work automatically (or at least if forced), regardless of whether the html document is clicked.
  
  
**Current behaviour**
When the app's window is focused, they keyboard input will work only after clicking inside the app. Otherwise it fails silently.
Forcing focus doesn't work.
  
  
[YouTube - quick test clicking the desktop and the app](https://youtu.be/y0ZEThHCLaE)
Also, notice how the app doesn't have keyboard focus if I click on the app's taskbar icon and then inside the app.
  
  
**Specifications**  
- OS: Windows 11 Pro x64 (no WSL)
- Neutralinojs version: v6.4.0
- Neutralinojs client library version: v6.4.0
- Neutralinojs CLI version: v11.6.0
  

  
**Additional context**
Why is this important? These kinds of apps are blocked by this bug:
1. Omnibars. Meaning floating bars like Spotlight/Alfred. The user starts typing their query as soon as the app is activated it. Requiring a click inside the app breaks the experience.
2. Custom terminals or apps with search bars that assign instant keyboard focus.
3. Apps that have custom keyboard shortcuts inside to launch functions.

Sahityaaryan:
## Description
  Fix keyboard input being ignored on Windows when the app window is refocused via the taskbar icon or title bar, without clicking inside the HTML document.

  When the Win32 parent window received `WM_ACTIVATE`, the handler dispatched the `windowFocus` event to JavaScript but never transferred keyboard focus to
  the WebView2 controller. The WebView2 `Chrome_WidgetWin_01` child control requires an explicit `MoveFocus()` call to accept keyboard input.

  Fixes #1491 

  ## Changes proposed

  - Added `focus()` method to `edge_chromium` class that calls `MoveFocus(COREWEBVIEW2_MOVE_FOCUS_REASON_PROGRAMMATIC)` on the WebView2 controller
  - Updated `WM_ACTIVATE` handler to call `focus()` when the window is activated, so the WebView2 control receives keyboard focus automatically

  ## How to test it

  - Create a new Neutralino app on Windows
  - Add a `keyup` event listener: `document.addEventListener("keyup", (e) => alert(e.key))`
  - Click outside the app to lose focus
  - Click the app's taskbar icon or title bar (NOT the HTML document) to refocus
  - Press a key — verify it registers without needing to click inside the document

  ## Next steps

  None.

  ## Deploy notes

  None.
  
  ## Demo Video
  
  https://drive.google.com/file/d/1mG4zejkuJpMNw_UE6L7rUSimlIB4vUDU/view?usp=sharing

### DIFF
--- a/lib/webview/webview.h
+++ b/lib/webview/webview.h
@@ -863,6 +863,13 @@ public:
     m_controller->put_Bounds(bounds);
   }
 
+  void focus(){
+    if(m_controller == nullptr){
+      return;
+    }
+    m_controller->MoveFocus(COREWEBVIEW2_MOVE_FOCUS_REASON_PROGRAMMATIC);
+  }
+
   void navigate(const std::string url) {
     auto wurl = to_lpwstr(url);
     m_webview->Navigate(wurl);
@@ -991,8 +998,10 @@ public:
               if(!windowStateChange) break;
               if(LOWORD(wp) == WA_INACTIVE)
                 windowStateChange(WEBVIEW_WINDOW_BLUR);
-              else
+              else{
                 windowStateChange(WEBVIEW_WINDOW_FOCUS);
+                w->m_browser->focus();
+              }
               break;
             case WM_DESTROY:
               PostQuitMessage(processExitCode);
--- a/lib/webview/webview.h
+++ b/lib/webview/webview.h
@@ -863,13 +863,6 @@ public:
     m_controller->put_Bounds(bounds);
   }
 
-  void focus(){
-    if(m_controller == nullptr){
-      return;
-    }
-    m_controller->MoveFocus(COREWEBVIEW2_MOVE_FOCUS_REASON_PROGRAMMATIC);
-  }
-
   void navigate(const std::string url) {
     auto wurl = to_lpwstr(url);
     m_webview->Navigate(wurl);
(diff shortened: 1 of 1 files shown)


shalithasuranga (approved):
Thanks so much :tada:

PR #1569, opened 2026-02-25
Fix extension processes becoming orphans on app exit

### LINKED ISSUE #1567: Extension processes become orphans when the app exits
When extensions are configured with a command, the spawned process handle is used only to send stdin then discarded. The loadedExtensions structure only stores string IDs so there is no way to track which process backs which extension.

When the app exits, none of those processes receive a kill signal. They keep running in the background as orphans, which is the root cause of the behavior reported in issues 1242 and 1541.

The fix needs to store the virtualPid alongside the extension ID, implement a cleanup function that kills each tracked process, and call it from the app exit handler.

nagadevikona20-max:
Closes neutralinojs/neutralinojs#1567

The core problem was in extensions_loader.cpp. When an extension has a command configured, spawnProcess returns a virtualPid that was only used to send stdin and then thrown away. loadedExtensions was a plain vector of strings with no process handle or pid attached to it. When the app exited nothing signaled those processes so they kept running as orphans.

The fix tracks the virtualPid alongside the extension id in a LoadedExtension struct. A cleanup() function was added that iterates over all loaded extensions and calls updateSpawnedProcess with exit for any that have a valid pid. The app exit handler now calls extensions::cleanup() before stopping the server so all extension processes are terminated cleanly.

getStats() in the extensions controller was also updated to return pid alongside the extension id so callers can know which system process backs each extension.

Files changed: extensions_loader.h, extensions_loader.cpp, api/extensions/extensions.cpp, api/app/app.cpp

### DIFF
--- a/api/app/app.cpp
+++ b/api/app/app.cpp
@@ -19,6 +19,7 @@
 #include "api/window/window.h"
 #include "api/os/os.h"
 #include "api/events/events.h"
+#include "extensions_loader.h"
 
 using namespace std;
 using json = nlohmann::json;
@@ -26,6 +27,7 @@ using json = nlohmann::json;
 namespace app {
 
 void exit(int code) {
+    extensions::cleanup();
     if(neuserver::isInitialized()) {
         neuserver::stop();
     }
--- a/api/extensions/extensions.cpp
+++ b/api/extensions/extensions.cpp
@@ -61,7 +61,16 @@ json broadcast(const json &input) {
 json getStats(const json &input) {
     json output;
     json stats;
-    stats["loaded"] = extensions::getLoaded();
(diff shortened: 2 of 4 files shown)


nagadevikona20-max:
HI @prasu-baran ,
Thanks for the suggestions!
I’ve added the duplicate extension load check in loadOne() and initialized virtualPid to -1 in the struct to avoid undefined behavior. please go through it !
Thankyou!

shalithasuranga:
Thanks so much for sending this. We need to look for some scenarios like interrupts, other processes initiated using Neutralinojs APIs. I'll re-open this later if we implement the same solution suggested from this PR :)

PR #1570, opened 2026-02-25
Add home and desktop to os.getPath()

### LINKED ISSUE #1568: os.getPath() has no support for home and desktop paths
os.getPath() covers most standard directories but is missing two of the most commonly needed paths: the user home directory and the desktop.

The sago library already exposes getDesktopFolder() but it was never wired up. For home, the platform specific approach is getpwuid(getuid())->pw_dir on Linux and macOS and SHGetKnownFolderPath(FOLDERID_Profile) on Windows.

Right now developers have to use os.getEnv("HOME") as a workaround which is not reliable cross platform. Both paths should be first class entries in getPath() under the names "home" and "desktop".

nagadevikona20-max:
Closes neutralinojs/neutralinojs#1568

os.getPath() was missing home and desktop, two paths that almost every desktop app needs at some point. This meant developers had to fall back to os.getEnv("HOME") which is not portable across platforms.

Desktop was straightforward since sago::getDesktopFolder() already exists in the bundled platform folders library, it just was not wired up.

For home, the implementation uses getpwuid(getuid())->pw_dir on Linux and macOS because reading the HOME env var can be wrong when the process was started under sudo or by a launcher that does not set it correctly. On Windows it uses SHGetKnownFolderPath(FOLDERID_Profile) which is the correct modern API.

Both are now available as "home" and "desktop" via os.getPath() on all three platforms.

Files changed: api/os/os.cpp

### DIFF
--- a/api/os/os.cpp
+++ b/api/os/os.cpp
@@ -23,6 +23,8 @@
 
 #if defined(__linux__) || defined(__FreeBSD__) || defined(__APPLE__)
 #include <unistd.h>
+#include <pwd.h>
+#include <sys/types.h>
 extern char **environ;
 #endif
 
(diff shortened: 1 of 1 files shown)


shalithasuranga (approved):
Thanks so much for your contribution @nagadevikona20-max :tada:

PR #1573, opened 2026-02-26
Fix exception slicing by catching std::exception by const reference

### LINKED ISSUE #1572: Fix object slicing by catching exceptions by const reference
### Describe the bug

Exceptions are caught by value instead of by reference in 6 places across the codebase. This causes object slicing the detailed error info from derived exceptions gets lost when they're copied into a base std::exception.

### To Reproduce

1. Open any of the 6 files listed below
2. Look at the catch blocks
3. They all use catch(exception e) instead of catch(const std::exception& e)

Affected files:
- settings.cpp 
- resources.cpp
- server/router.cpp 
- server/neuserver.cpp 
- api/window/window.cpp

### Expected behavior

All catch blocks should use catch(const std::exception& e) to avoid slicing and keep the full error message. The project already does this correctly at neuserver.cpp line 188 so it's just about being consistent.

N/A

### Specifications

### Additional context

Quick one-line fix per site. Just change catch(exception e) to catch(const std::exception& e). No side effects.

Tahanyemad16:
## Description

All 6 `catch` blocks in the codebase were catching `std::exception` by value (`catch(exception e)`), which causes object slicing. Derived exception data (like `json::parse_error` position info) gets silently lost. Changed them to catch by const reference.

## Changes proposed

- Replaced `catch(exception e)` with `catch(const exception& e)` in 6 locations across 5 files

## How to test it

- Build the project: `cd build && ninja`
- Run `grep -rn "catch(exception e)" --include="*.cpp" --exclude-dir=lib .` to confirm zero matches remain
- Run the test suite: `cd spec && npm test`

## Next steps

None.

## Deploy notes

None.

Close #1572

### DIFF
--- a/api/window/window.cpp
+++ b/api/window/window.cpp
@@ -313,7 +313,7 @@ bool __loadSavedWindowProps() {
         SetWindowPlacement(windowHandle, &wp);
         #endif
     }
-    catch(exception e) {
+    catch(const exception& e) {
         debug::log(debug::LogTypeError, errors::makeErrorMsg(errors::NE_CF_UNBLWCF, string(NEU_WIN_CONFIG_FILE)));
         return false;
     }
--- a/resources.cpp
+++ b/resources.cpp
@@ -129,7 +129,7 @@ bool __makeBundleFileTree() {
     try {
         files = json::parse(headerContent);
     }
-    catch(exception e) {
+    catch(const exception& e) {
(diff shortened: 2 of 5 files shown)


shalithasuranga (approved):
Thanks so much @Tahanyemad16 for your contribution :tada:

Issue #1080, opened 2023-03-07
`window.setSize` crashes on Windows


shalithasuranga:

shalithasuranga:
Linking this issue to the main #648 :tada:

PR #1577, opened 2026-02-27
fix: re-enable full window.setSize tests for all platforms

### LINKED ISSUE #1566: Re-enable full window.setSize tests for Windows in spec/window.spec.js
Steps to Reproduce / Expected Behavior

Locate the following TODO comment around line 291 in 

spec/window.spec.js : // TODO: Remove this check after fixing: #1080
Remove the if(process.platform == 'win32') conditioning structure.
Ensure that the describe('window.setSize'...) block that was previously inside the else clause (which contains the "works without throwing errors" test) is used universally, without conditional branching.

VibhuGupta-dev:
## Description
Removed the platform-specific `if/else` conditioning for `window.setSize` tests and the leftover TODO comment. The underlying Windows crash bug (#1080) was fixed in November 2023, but the conditional test block was never cleaned up.

## Changes proposed
- Removed `if(process.platform == 'win32')` conditional block from `window.setSize` tests
- Removed TODO comment referencing issue #1080
- Full `window.setSize` tests now run universally on all platforms including Windows

## How to test it
- Run the window spec tests on Windows and verify `window.setSize` tests pass without platform-specific conditioning

## Next steps
Closes #1566
Relates to #1080

## Deploy notes
None.

### DIFF
--- a/spec/window.spec.js
+++ b/spec/window.spec.js
@@ -288,44 +288,31 @@ describe('window.spec: window namespace tests', () => {
         });
     });
 
-    // TODO: Remove this check after fixing: #1080
-    if(process.platform == 'win32') {
     describe('window.setSize', () => {
-        it('exports the function to the app', async () => {
-            runner.run(`
-                await __close(typeof Neutralino.window.setSize);
-            `);
-            assert.equal(runner.getOutput(), 'function');
-        });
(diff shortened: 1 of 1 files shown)


shalithasuranga (approved):
Thanks so much :tada:

PR #1579, opened 2026-02-27
fix(fs): use atomic counter for opened file IDs to prevent collisions

### LINKED ISSUE #1578: filesystem.openFile() assigns duplicate file IDs after closing files
**Describe the bug**
openFile() uses openedFiles.size() to generate virtual file IDs. When a file is closed and erased from the map, the size shrinks, causing new files to receive IDs that collide with existing open files. This silently overwrites file handles, leading to data corruption and memory leaks.

**To Reproduce**
1. Open 3 files → they get IDs 0, 1, 2 (map size = 3)
2. Close the file with ID 1 → map size drops to 2
3. Open a new file → it gets ID 2 (because size = 2)
4. Now two different file handles share ID 2 → **data corruption**

**Expected behavior**
Each openFile() call should return a unique ID that is never reused, regardless of prior closes.

If applicable, add screenshots/screen recordings to help explain your problem.

**Specifications**  
 - OS: All (platform-independent logic bug)
 - Affects: api/fs/fs.cpp, function fs::openFile()

abhi0324:
## Description

Fixes #1578. `openFile()` used `openedFiles.size()` to assign virtual file IDs, which caused ID collisions when files were closed and new ones opened. Closing a file erases it from the map, shrinking the size, so subsequent opens could produce an ID already held by another open file, silently overwriting its handle.

## Changes proposed

- Replaced `openedFiles.size()` with a monotonically increasing `atomic<int>` counter (`nextFileId`) so IDs are never recycled
- Added `#include <atomic>` for `std::atomic` support

## How to test it

- Open 3 files via `Neutralino.filesystem.openFile()` (IDs 0, 1, 2)
- Close the middle file (ID 1) via `updateOpenedFile(1, 'close')`
- Open a new file - it should get ID 3, not 2
- Verify reads/seeks on the file with ID 2 still operate on the correct original file

## Next steps

None.

## Deploy notes

None.

### DIFF
--- a/api/fs/fs.cpp
+++ b/api/fs/fs.cpp
@@ -6,6 +6,7 @@
 #include <algorithm>
 #include <cctype>
 #include <filesystem>
+#include <atomic>
 
 #if defined(__linux__) || defined(__APPLE__) || defined(__FreeBSD__)
 #include <unistd.h>
@@ -41,6 +42,7 @@ using json = nlohmann::json;
 
 map<int, ifstream*> openedFiles;
 mutex openedFilesLock;
+atomic<int> nextFileId{0};
 efsw::FileWatcher* fileWatcher;
 map<efsw::WatchID, pair<efsw::FileWatchListener*, string>> watchListeners;
 mutex watcherLock;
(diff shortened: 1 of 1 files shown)


abhi0324:
Hey @shalithasuranga, please review

abhi0324:
@SharonIV0x86 Thanks for pointing this out. I reviewed #1552 and it addresses the same issue regarding file ID reuse after closing files.
Since that PR already handles the fix, I’ll close this one to avoid duplication.
Appreciate the clarification.

PR #1582, opened 2026-02-28
Fix app crashes on macOS when icon path has no leading slash

### LINKED ISSUE #1344: Crash on macOS if icon path doesn't start with a `/`
**Describe the bug**
`neu run` crashes if the link to an icon set in `neutralino.config.json`, `modes.window.icon` doesn't start with a /.

* `resources/icon.png` crashes.
* `./resources/icon.png` crashes.
* `/resources/icon.png` works.

The crashes don't happen on Windows at all. I don't know what happens ion Linux.
This can be replicated with a default template created by `neu create sample-app`. 

**To Reproduce**
1. `neu create sample-app`
2. `cd ./sample-app`
3. `neu run` — Works, ok!
4. Change `neutralino.config.json`:
```diff
- "icon": "/resources/icons/appIcon.png"
+ "icon": "resources/icons/appIcon.png"
```
5. Run `neu run` again — observe a crash.

**Expected behavior**
Windows and macOS configuration behavior must match. In one way or another. Preferably all three path styles (absolute-ish, relative, relative starting with `./`) should be supported.

изображение
изображение

**Specifications**  
 - OS: macOS Monterey (12.7.6)
 - Neutralinojs version: 5.5.0
 - Neutralinojs client library version: 5.5.0
 - Neutralinojs CLI version: 11.3.0

**Additional context**
—

nagadevikona20-max:
Description
Closes #1344 
Setting an icon path like "resources/icons/appIcon.png" or "./resources/icons/appIcon.png"
in neutralino.config.json crashes the app on macOS. The resource bundle lookup expects
paths to begin with "/", so without it getFile returns empty data. Passing that empty
data into NSImage initWithData: returns nil, and calling setApplicationIconImage: with
nil throws an uncaught NSException. Windows is more forgiving about this which is why
the same config works fine there.

Changes proposed
 - Normalize the icon path before resource lookup so "resources/x.png" and
   "./resources/x.png" are both treated the same as "/resources/x.png".
 - Return early with a logged error if the file still cannot be found, instead of
   passing empty bytes into the platform icon APIs.

How to test it
 - Set "icon": "resources/icons/appIcon.png" in neutralino.config.json (no leading slash).
 - Run the app on macOS and confirm it no longer crashes and the icon loads correctly.
 - Repeat with "./resources/icons/appIcon.png" and confirm the same result.
 - Confirm that "/resources/icons/appIcon.png" still works as before on all platforms.

Next steps
None.

Deploy notes
None.

### DIFF
--- a/api/window/window.cpp
+++ b/api/window/window.cpp
@@ -876,7 +876,20 @@ void exitFullScreen() {
 }
 
 void setIcon(const string &iconFile) {
-    fs::FileReaderResult fileReaderResult = resources::getFile(iconFile);
+    string normalizedIconFile = iconFile;
+    if(!normalizedIconFile.empty() && normalizedIconFile[0] != '/') {
+        if(normalizedIconFile.size() > 1 && normalizedIconFile[0] == '.' && normalizedIconFile[1] == '/') {
+            normalizedIconFile = normalizedIconFile.substr(1);
+        }
+        else {
+            normalizedIconFile = "/" + normalizedIconFile;
+        }
+    }
+    fs::FileReaderResult fileReaderResult = resources::getFile(normalizedIconFile);
+    if(fileReaderResult.status != errors::NE_ST_OK) {
+        debug::log(debug::LogTypeError, errors::makeErrorMsg(errors::NE_RS_NOPATHE, normalizedIconFile));
+        return;
+    }
     string iconDataStr = fileReaderResult.data;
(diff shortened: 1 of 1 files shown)


shalithasuranga:
Thanks so much for sending this. In Neutralinojs, `/path/..` is a standard since these resources are mainly loaded from a bundle in production. We need to display a top-level general error if `/` is not present, and we'll look into it :) Thanks

Issue #1409, opened 2025-04-08
Windows 10 or MacOS App silently does not start. No log. Nothing.


uschiklo:
**Describe the bug**
Windows 10 or MacOS App silently does not start. No log. Nothing.

**To Reproduce**
1. neu build with 6.0.0
2. app starts on windows 11
3. copy app to windows 10, does not start without error
4. same with mac os.

**Expected behavior**
App starts.

**Specifications**  
 - OS: Windows 10 x64, MacOS
 - Neutralinojs version: 6.0.0
 - Neutralinojs client library version: 5.6.0
 - Neutralinojs CLI version: 11.4

**Additional context**
The same Problem was also with pre 6.0.0

hschneider:
Regarding macOS: 
If you copy the app to a new machine, you might need to clear Gatekeeper's quarantine flag, before you can launch the app:
`sudo xattr -d -r com.apple.quarantine appname`

If the error persists:
Can you please provide the macOS version and the steps to reproduce?

uschiklo:
Thank you for the fast answer. Ill try it the next days. My brother will test it on his MacOS. 

Meanwhile i installed Lubuntu and also have the same problem. Linux version of my App does not start and no Error or Info is shown.

I also cant find any info at neutralino Docs how WebView is used in linux or MacOS. 
Neutralino js is really easy to use as a dev in windows 11 but having so much problems distributing there is little benefit of neutralino js.

uschiklo:
> Regarding macOS: If you copy the app to a new machine, you might need to clear Gatekeeper's quarantine flag, before you can launch the app: `sudo xattr -d -r com.apple.quarantine appname`
> 
> If the error persists: Can you please provide the macOS version and the steps to reproduce?

My brother tested the file for MacOS 15 ARM. 
He set your sudo bash and MacOS errors with"... can't be opened."

i also tried your script to bundle the apps for mac or linux. Can you please describe a better tutorial? Its really hard to get the script running. Do i need to setup your script completly in linux setup? In Windows 11 and WSL i failed.

hschneider:
You should be able to compile the app and start it right from the terminal, without the bundling-script. 
This works here on macOS 15 Intel. 

The app bundle script will only work, if the compiled Nautralino app itself works.

On macOS ARM, there is a problem with the neutralino binary itself.
You can solve it by copying `bin/neutralino-mac_x64` to `bin/neutralino-mac_arm64`.
[...]

Issue #959, opened 2022-07-20
Apps are not natively supported by Windows 10. 


kolbasa:
**Describe the bug**
Apps cannot be used with a fresh Windows 10 installation.
Only an empty window opens. No error message, no log messages.
The problem: Windows 10 does not come bundled with WebView2.
Even if all Windows updates are installed.

**To Reproduce**
1. Create a fresh Windows 10 Install
2. Open a simple Neutralinojs App

**Expected behavior**
Ideally, the first launch would prompt the user to install WebView2.
Or at least there should be a meaningful error message.
Other similar frameworks install WebView2 automatically (e.g. Tauri).

**Screenshots/Screen recordings**
![Bildschirmfoto 2022-07-20 um 11 24 01](https://user-images.githubusercontent.com/3901582/180023345-e368f069-bdbf-4916-b608-8784be67ed38.png)

**Specifications**  
-OS: Windows 10 x64
-Neutralinojs version: v4.6.0
-Neutralinojs client library version: v3.5.0
-Neutralinojs CLI version: v9.3.1

ZaDarkSide:
Yeah same here, it should at least prompt the user to install WebView2

shalithasuranga:
Closing this since now Neutralinojs displays a message if webview2 is not installed. Thanks for reporting this :tada:

PR #1585, opened 2026-03-01
fix: added stderr error output for silent startup failures

Sahityaaryan:
## Description

  When webview initialization fails (missing WebView2 on Windows or WebKit2GTK on Linux), the only error reporting was a GUI dialog (`pfd::message`) that
  depends on the very window system that's failing. This causes apps to silently exit with no log, no error, nothing — as reported in #1409 and previously
  in #959.

  This PR adds `fprintf(stderr, ...)` calls to all startup failure paths so errors are always visible when the app is launched from a terminal.

  ## Changes proposed

   - Added stderr output in `lib/webview/webview.h` when `dlopen` fails for WebKit2GTK (Linux) — includes `dlerror()` for the exact reason
   - Added stderr output in `lib/webview/webview.h` when `CreateCoreWebView2EnvironmentWithOptions` fails (Windows) — includes the HRESULT code
   - Added stderr fallback in `main.cpp` before every `pfd::message` + `std::exit(1)` path (webview init, server init, config parse)

  ## How to test it

   - Build the project and run the binary from a terminal on a system without WebKit2GTK (Linux) or WebView2 (Windows 10)
   - Verify the error message appears in stderr instead of a silent exit
   - On a system with all dependencies, verify the app starts normally with no extra output

  ## Next steps

  None.

  ## Deploy notes

  None.

## Demo of the changes:
[demo_of_changes_made](https://drive.google.com/file/d/1NoyI2h-x8b-lnhr5f_g4EtRnL5ihCdGP/view?usp=sharing)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ rename `Unreleased` topic with the new version tag. Finally, create a new `Unrel
 ## Unreleased
 
 ### API: computer
-- Implement `computer.getMousePosition(x, y)` to update the current mouse cursor position.
+- Implement `computer.setMousePosition(x, y)` to update the current mouse cursor position.
 - Implement `computer.setMouseGrabbing(grabbing; boolean)` to activate/deactivate confining the mouse cursor within the native app window. If `grabbing` is set to `true`, the mouse cursor always stays within the window boundaries, so this feature helps create interactive games and similar apps operated using the mouse.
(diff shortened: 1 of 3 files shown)


shalithasuranga:
Thanks for sending this. We would like to use a UI dialog since the error is for app users. Let's discuss more about logging by reopening this if more developers request logging :) Thanks

PR #1587, opened 2026-03-02
Fix crash caused by unguarded std::stoi() calls

### LINKED ISSUE #1583: Fix unguarded `stoi()` calls to prevent crashes on invalid inputs
### Describe the bug

`stoi()` calls in `settings.cpp` and `resources.cpp` are not wrapped in try-catch. If the input string is not a valid integer, `stoi()` throws `std::invalid_argument` or `std::out_of_range` and crashes the process.

- `settings.cpp` : passing `--port=abc` on the command line crashes the app.
- `resources.cpp` : a malformed `.asar` resource file with a non-numeric offset crashes the app.

### To Reproduce

1. Build the Neutralinojs binary
2. Run it with `./neutralino --port=abc`
3. The process crashes with an unhandled `std::invalid_argument` exception

### Expected behavior

The app should log an error and fall back to a default value or exit gracefully instead of crashing with an unhandled exception.

N/A

### Specifications

### Additional context

There are 3 unguarded `stoi()` calls:
- `settings.cpp`   `stoi(cfgOverride.value)` in config override parsing
- `resources.cpp`   `stoi(p.second)` in ` getFileFromBundle()`
- `resources.cpp`  `stoi(p.second)` in ` getFileFromEmbedded()`

subhu339:
Fixes #1583

## Description

This PR prevents application crashes caused by unguarded `std::stoi()` calls when invalid numeric values are provided (e.g., invalid CLI overrides or corrupted resource offsets).

## Changes proposed

- Wrapped numeric conversions in try–catch blocks in resources.cpp
- Added safe handling for CLI numeric overrides in settings.cpp

## How to test it

- Build the project
- Run the binary with an invalid value (--port=abc)
- Verify that the application does not crash

## Next steps

None.

## Deploy notes

None.


shalithasuranga:
Closed since the PR contains the malicious payload. Please check whether your computer is compromised. Thanks

PR #1591, opened 2026-03-03
Feat/move to trash 1031

### LINKED ISSUE #1031: "Move to trash"-functionality

Sahityaaryan:
## Description

  Adds `filesystem.moveToTrash(path)` API method that moves files and directories to the OS
  trash/recycle bin instead of permanently deleting them. Resolves #1031.

  ## Changes proposed

  - Added `moveToTrash` controller with platform-specific trash implementations (XDG Trash spec on
  Linux/FreeBSD, NSFileManager on macOS, SHFileOperationW on Windows)
  - Added `NE_FS_TRSERR` error code for trash operation failures
  - Registered `filesystem.moveToTrash` route and added spec tests

  ## How to test it

  - Run specs/tests
  - Create a test file via `Neutralino.filesystem.writeFile()`, then call
  `Neutralino.filesystem.moveToTrash()` on it and verify the file is no longer at the original path
  - On Linux, verify the file appears in `~/.local/share/Trash/files/` with a corresponding `.trashinfo`
   entry

  ## Next steps

  None.

  ## Deploy notes

  None.


shalithasuranga (approved):
Thanks so much. I've added without the Linux implementation (need to check further about a standard implementation, etc) :)

PR #1596, opened 2026-03-05
fix: prevent path traversal in resources.extractFile

### LINKED ISSUE #1595: Fix: Arbitrary File Write via `destination` in `res::extractFile`
## Describe the bug

The `res::controllers::extractFile()` function in `api/res/res.cpp` accepts a `destination` path directly from user-provided JSON input and passes it to `resources::extractFile()`, which creates directories and writes file content **without any path traversal sanitization**.

A malicious client connected via the WebSocket native bridge can supply a `destination` containing `../` sequences (e.g., `../../../tmp/pwned.html`) to write arbitrary files outside the application's working directory.  
This results in a **CWE-22: Path Traversal** vulnerability that allows arbitrary file writes.

## To Reproduce

1. Start a Neutralinojs app in **bundle mode** or **embedded mode** with a known resource file (e.g. `/index.html`).
2. Send a native WebSocket request using the `resources.extractFile` method.
3. Provide a `destination` path containing directory traversal sequences such as `../../../tmp/pwned.html`.
4. Observe that the file is extracted outside the application's working directory.
5. The server returns a successful response.

## Expected behavior

The controller should validate that the resolved `destination` path does not escape the application's working directory. If it does, the operation should be rejected with an error (e.g. `NE_RS_FILEXTF`).

## Specifications

- OS: Linux, macOS, Windows
- Neutralinojs version: latest main branch

Tahanyemad16:
## Description

Fixes #1595 — Path traversal vulnerability in `res::controllers::extractFile()`.

The `extractFile()` controller in `api/res/res.cpp` accepted the `destination` parameter from user-provided JSON input and passed it directly to filesystem write operations without any path boundary validation. A malicious client could supply `../` sequences in the destination to write files outside the application's working directory (CWE-22).

## Changes proposed

- Added path traversal sanitization in `res::controllers::extractFile()` using `filesystem::weakly_canonical()` to resolve the destination path and verify it stays within the current working directory
- If the resolved destination escapes `cwd`, the function now returns an `NE_RS_FILEXTF` error and refuses the operation

## How to test it

- Start a Neutralinojs app in bundle or embedded mode
- Send a `resources.extractFile` call with a traversal destination (e.g. `"destination": "../../../tmp/test.html"`) and verify it returns an `NE_RS_FILEXTF` error
- Send a valid destination within the app directory and verify extraction still succeeds
- Run existing specs/tests to confirm no regressions

## Next steps

- Apply the same path traversal validation to `res::controllers::extractDirectory()` which has the same vulnerability

## Deploy notes

None.

### DIFF
--- a/api/res/res.cpp
+++ b/api/res/res.cpp
@@ -1,4 +1,3 @@
-#include <iostream>
 #include <string>
 #include <vector>
 #include <filesystem>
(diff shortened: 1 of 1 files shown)


shalithasuranga:
Please note that we are trying to answer these issues with a scoped file system in the future. Thanks so much for sending your work. At this stage, we treat `resources` APIs as normal file handling APIs. Thanks

PR #1598, opened 2026-03-05
hostname shown

prasu-baran:
## Description

Adds os.getHostname() to retrieve the system hostname natively without spawning a shell.
## Changes proposed

 - Added getHostname() C++ controller in api/computer/computer.cpp using gethostname() on Linux/macOS/FreeBSD and GetComputerNameA() on Windows
 - Added <unistd.h> include for Linux to support gethostname()
 - Declared getHostname in api/computer/computer.h
 - Registered computer.getHostname in the router method map (server/router.cpp)

## Next steps

Documentation needs to be updated to show it's an added feature.
The official neutralino.js client library needs getHostname added to the computer namespace bindings


api/computer/computer.cpp
```diff
@@ -394,71 +197,32 @@ json getMousePosition(const json &input) {
     return output;
 }
 
-json setMousePosition(const json &input) {
+json getHostname(const json &input) {
```
shalithasuranga, line 413: Since this is a simple getter function, please avoid using an error message and return an empty string if the hostname cannot be retrieved. Also. Make sure to use `GetComputerName` (not the ASCII version) with wstring (using `helper::..` like in other places). Thank you

shalithasuranga (changes requested):

api/computer/computer.cpp
prasu-baran: okay got it will do accordingly

shalithasuranga:
Thanks so much for updating this @prasu-baran. Please fix merge conflicts (sync with the upstream), then we can merge this :)

shalithasuranga (approved):
Looks great. Thanks for updates :)

PR #1603, opened 2026-03-05
Fix use app data path for Chrome mode user-data-dir (#1413)

### LINKED ISSUE #1413: External URLs don't persist their session
**Describe the bug**
This is a direct reference to https://github.com/neutralinojs/neutralinojs-cli-appify/issues/19

I tried to create a desktop app from a web app, but every time I open the binary, I'm requested to log in again. There's no clear information if there's anything extra to be done to allow the internal browser to persist the user session.

**To Reproduce**
1. appify an external URL (i.e. use some publicly accessible system on `url` from `neutralino.config.json`)
2. run the binary
3. wait for the load, and log in
4. close the application
5. run the binary again

**Expected behavior**
You're still logged in

**What happens**
You see the log in screen again

**Specifications**  
 - OS: Linux Mint 22 with Cinnamon DE
 - Neutralinojs version: 6.0.0
 - Neutralinojs client library version: 6.0.0
 - Neutralinojs CLI version: 11.4.0

nagadevikona20-max:
Description
Fixes #1413 

When running a Neutralinojs app in chrome mode, browser session data (cookies, login state) was not persisting between launches. The --user-data-dir was being set to <appPath>/.tmp/chromedata, which places the chrome profile folder next to the binary. For appified apps the binary often lives in a location that is read-only or gets replaced on update, so the session data either could not be written or was simply treated as throwaway. The .tmp folder name itself signals that this was never meant to be permanent storage.

Changes proposed

- Changed the chrome user-data-dir from joinAppDataPath("/.tmp/chromedata") to joinAppDataPath("/chromedata") so it now respects the dataLocation config setting the same way the rest of the framework does
- Updated the JSON schema description to reflect the corrected path

How to test it

- Set mode to chrome and url to any external site that requires login in neutralino.config.json
- Build and run the binary, log in to the site
- Close and relaunch the binary
- You should remain logged in
- Set dataLocation: "system" in the config and repeat, confirm session still persists and data lands under the OS user data directory

Next steps

None.

Deploy notes

On first launch after upgrading, the old .tmp/chromedata directory will be automatically moved to chromedata if it exists and the new path does not, so existing sessions are preserved.


SagarJadhav30 (commented):
best work

nagadevikona20-max:
@SagarJadhav30 Thank you, I’m glad it met expectations

shalithasuranga:
Closing since apps won't always use `.tmp` for storage and can use the system-specific data directory using `dataLocation` config option. Thanks so much for sending this :)

PR #1606, opened 2026-03-06
Fix check image format before text in clipboard.getFormat() (#1268)

### LINKED ISSUE #1268: `getFormat()` return an empty string even when the clipboard contains an image
**Describe the bug**
The `getFormat()` function should return 'image' when the clipboard contains an image but it returns an empty string

**To Reproduce**
1.  run `cd spec && npm test clipboard`

**Expected behavior**
It should return 'image'

If applicable, add screenshots/screen recordings to help explain your problem.

**Specifications**  
 - OS: Windows
 - Neutralinojs version: v5.2.0
 - Neutralinojs client library version: v5.2.0
 - Neutralinojs CLI version: v11.2.0

*Note: you can enter the `neu version` command to find these details*

**Additional context**

nagadevikona20-max:
Description
Closes #1268
clipboard.getFormat() may return "text" when the clipboard actually contains an image. On Windows, copying an image also creates a text-compatible clipboard entry, and since text_format() was checked first it matched before image_format().

This PR reorders the checks so image is evaluated before html and text.

---

Changes proposed

Reorder format checks in getFormat() so image_format() is evaluated before html_format() and text_format().

---

How to test it

Copy an image to the clipboard and run Neutralino.clipboard.getFormat().  
It should return "image".

---

Next steps
None.

---
Deploy notes
None.


shalithasuranga (approved):
Thanks so much :tada:

PR #1615, opened 2026-03-06
fix: add missing unNormalizePath call in showOpenDialog

itssagarK:
## Description
`showOpenDialog` was missing a call to `helpers::unNormalizePath()` for the `defaultPath` parameter, while both `showSaveDialog` and `showFolderDialog` already had this call.

This means on Windows, if a `defaultPath` with forward slashes was passed to `showOpenDialog`, the dialog could open in the wrong directory.

## Changes proposed
- Added `defaultPath = helpers::unNormalizePath(defaultPath)` in `showOpenDialog` to match existing behavior in `showSaveDialog`  and `showFolderDialog`
 
## How to test it
- Call `Neutralino.os.showOpenDialog` with a `defaultPath` on Windows
- Verify the dialog opens in the correct directory

image


shalithasuranga (approved):
Thanks so much :tada:

PR #1616, opened 2026-03-06
fix: added desktop path support to os.getPath()

itssagarK:
## Description
`os.getPath()` was missing support for the `desktop` path name. Calling `os.getPath('desktop')` would throw `NE_OS_INVKNPT` error.

The `sago` library already provides `getDesktopFolder()` but it was never added to the if-else chain in `os::getPath()`.

## Changes proposed
- Added `desktop` case to `os::getPath()` using `sago::getDesktopFolder()`

## How to test it
- Call `await Neutralino.os.getPath('desktop')`
- Verify it returns the correct Desktop folder path
image


shalithasuranga:
Thanks so much for sending this. Closing this since the same feature was added by a previous PR.

itssagarK:
Understood! Thanks for letting me know. I'm glad to see the feature was added either way

PR #1625, opened 2026-03-08
fix: free websocket server instance in stop() to prevent memory leak

Shashwat325:
**Issue**
The websocket server object created with \`new\` in init() was never deleted in stop(). This causes memory leaks if init() is called multiple times (e.g., during server restart) or in Cloud mode where the process runs persistently.

**Changes**
Added ```delete server;
    server = nullptr;```
to ```void stop()```

 **Result**
This will free heap memory after the server stops preventing memory leaks.

### DIFF
--- a/server/neuserver.cpp
+++ b/server/neuserver.cpp
@@ -165,6 +165,8 @@ void startAsync() {
 
 void stop() {
     server->stop_listening();
+    delete server;
+    server = nullptr;
 }
 
 void handleMessage(websocketpp::connection_hdl handler, websocketserver::message_ptr msg) {


prasu-baran:
Hello @Shashwat325 ! ,
Just a suggestion for you while making a new branch for a new PR kind go to the main branch and them pull the original code at the github thereafter make a new branch and open the new PR . Your previous commits are also coming here of const and map which you did earlier . 
Have a look over it too.
Thanks !

prasu-baran:
Yes definitely it should be cleaned and this issue should solve one particular thing only as it gives clarity about the current issue being solved .

PR #1607, opened 2026-03-06
implements network interface api 

prasu-baran:
## Description

Adds computer.getNetworkInterfaces() API to retrieve network interface info (name, IPv4, IPv6, MAC, isUp, isLoopback) with an optional excludeLoopback parameter.
## Changes proposed

 - Added getNetworkInterfaces controller in api/computer/computer.cpp with platform-specific implementations (getifaddrs on Linux/macOS/FreeBSD, GetAdaptersAddresses on Windows)
 - Registered the method in server/router.cpp and added iphlpapi/ws2_32 to Windows linker libs in CMakeLists.txt
 

## How to test it

 - Run spec/computer.spec.js — 3 new tests covering return shape, excludeLoopback filtering, and count consistency

## Next steps

None.

## Deploy notes

None.

Solves #1604

PR #1629, opened 2026-03-09
fix: improve test suite reliability and cross-platform support

itssagarK:
While going through the CI failure logs on PR #1607, I noticed all three network interface tests were failing on Windows with `NL_SP_MAXTIMT` instead of real output. I traced this back to spec/runner.js and found the root cause , the app-level timeout was hardcoded to 20 seconds for all platforms. Windows CI runners take longer to respond for certain APIs, so the timeout fires before the result is written.

While fixing that, I found the same issue in spec/index.js (Mocha timeout), and a separate bug in spec/os.spec.js where a missing `await` was causing the os.open error test to hang indefinitely.

## Changes made

**spec/runner.js**
- Added `DEFAULT_TIMEOUT` constant — 40s on Windows, 20s on Linux/Mac
- Injected via `{TIMEOUT}` placeholder in `makeAppSource()` so the value reaches the Neutralino app as a plain number (the template runs in a separate context from Node.js and cannot access variables directly)
- Windows binary was hardcoded to `win_x64.exe` — changed to use `process.arch` dynamically, same as Linux and Mac
- Added proper error throw for unsupported platforms instead of silently producing a broken command string
- Replaced `==` with `===` for all platform string comparisons

**spec/index.js**
- Mocha timeout was also hardcoded to 20s — applied the same platform-aware fix

**spec/os.spec.js**
- `os.open` error test was missing `await` on `__close(error.code)` in the catch block — without it the async file write never completes and the app hangs until timeout fires

## How to test it

- Run `npm run test app` — 14 passing, 0 failing
image

- Run `npm run test os` — 45 passing, 0 failing
image

## Next steps

None.

## Deploy notes

None.

## Related

Timeout issue identified through CI failure in #1607. This fix should help any future API additions pass Windows CI reliably.

Note: `os.execCommand` fails intermittently on Windows regardless of these changes — confirmed pre-existing flaky behavior by testing with and without this fix.

### DIFF
--- a/spec/index.js
+++ b/spec/index.js
@@ -11,7 +11,8 @@ fs.readdirSync(testDir).filter((file) => file.includes(specModule + '.spec.js'))
       mocha.addFile(path.join(testDir, file));
 
 });
-mocha.timeout(20000);
+const MOCHA_TIMEOUT = process.platform === 'win32' ? 40000 : 20000;
+mocha.timeout(MOCHA_TIMEOUT);
 mocha.run((failures) => {
     process.exitCode = failures ? 1 : 0;
 });
--- a/spec/os.spec.js
+++ b/spec/os.spec.js
@@ -532,7 +532,7 @@ describe('os.spec: os namespace tests', () => {
                 try {
                     await Neutralino.os.open(12345);
                 } catch (error) {
-                    __close(error.code);
+                    await __close(error.code);
                 }
             `);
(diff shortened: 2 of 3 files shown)


shalithasuranga:
Thanks so much for sending this. Seems like the current timeout works fine. I'll reopen this if there are issues after merging #1607

PR #1632, opened 2026-03-10
feat: add os.getUserInfo() API and fix missing os.getPath router registration

itssagarK:
Adds a new os.getUserInfo() native API and fixes a silent bug where os.getPath() was never reachable from the frontend.

## What I built

**os.getUserInfo() — new native API**

Returns the current system user's information. I used `getpwuid(getuid())` on Linux/Mac via `<pwd.h>` which reads directly from the system user database, and `GetUserName()` + `USERPROFILE` on Windows.

The response object contains:
- `username` — logged-in user's name
- `homeDir` — user's home directory  
- `shell` — default shell on Linux/Mac, cmd path on Windows

## Bug I found while doing this

While adding `os.getUserInfo` to `server/router.cpp`, I noticed that `os.getPath` — which was implemented in PR #1616 — was never registered in the router's method map. Every frontend call to `Neutralino.os.getPath()` would silently fail with a method-not-found error even though the C++ implementation existed. Added the missing entry.

## Files changed

- `api/os/os.h` — declared os::getUserInfo() and its controller
- `api/os/os.cpp` — cross-platform implementation + JSON controller  
- `server/router.cpp` — registered os.getUserInfo and fixed os.getPath

## How to test it

Build compiles clean on Windows x64 — 2/2 files, zero errors. 
image

### DIFF
--- a/api/os/os.cpp
+++ b/api/os/os.cpp
@@ -682,6 +682,15 @@ json setTray(const json &input) {
 
     if(helpers::hasField(input, "menuItems")) {
         int menuCount = input["menuItems"].size();
+
+        // Free any previously allocated menu items beyond the new count
+        for(int j = menuCount; j < NEU_MAX_TRAY_MENU_ITEMS; j++) {
+            if(menus[j].id == nullptr && menus[j].text == nullptr) break;
+            delete[] menus[j].id;
+            delete[] menus[j].text;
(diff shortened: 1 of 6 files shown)


shalithasuranga:
Closing this since this will be a part of the ongoing GSoC 2026 project. Thanks so much for your work :tada:

PR #1633, opened 2026-03-10
feat: add computer.getBatteryInfo API

Shashwat325:
Adds a new getBatteryInfo().
**Changes**
api/computer/computer.cpp — added ```getBatteryInfo()``` implementation for all 3 platforms.
api/computer/computer.h — added ```getBatteryInfo()``` declaration to the controllers namespace.
spec/computer.spec.js — added 3 new spec tests covering type checking, valid range validation, and no-battery edge case for desktop machines.
servers/router.cpp— added ```{"computer.getBatteryInfo", &computer::controllers::getBatteryInfo}``` to methodMap
**Returns**
- available: whether battery info could be read
- charging: whether the device is plugged in
- level: battery percentage (0-100), or -1 if unavailable
**Test**
Ran full test using ```npm test computer``` command
Screenshot 2026-03-10 180559
all test cases of specs passing

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -148,22 +148,20 @@ if(WIN32)
     _UNICODE
   )
 elseif(APPLE)
-    # macOS
-    find_library(WebKit_FRAMEWORK WebKit)
-    find_library(COCOA_FRAMEWORK Cocoa)
-    find_library(SCREENCAPTUREKIT_FRAMEWORK ScreenCaptureKit)
-    find_library(IOKIT_FRAMEWORK IOKit)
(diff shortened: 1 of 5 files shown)


prasu-baran:
Hi @Shashwat325 ,

I think your linux implementation needs little more work right now you assume Bat0 and Bat1. But some system may use Bat2 or CMB0 like that . So have a look over it too
kIOPSPowerSourceStateKey : This actually checks AC power, not charging. So work over it too .

 So have a look over it rest implementation is great .

shalithasuranga:
Closing this since this will be a part of the ongoing GSoC 2026 project. Thanks so much :)

PR #1636, opened 2026-03-14
fix(linux): prevent Gdk-CRITICAL gdk-monitor-get-workarea assertion on missing monitor data

uv-goswami:
## Description
This PR fixes a Gdk-CRITICAL assertion error on Linux that occurs when the framework attempts to center the window in environments where a primary monitor is not accessible 

## Changes proposed
 - Added null pointer checks for  **GdkDisplay**  and **GdkMonitor** objects in window centering logic.
 - Added fallback to default window coordinates to (0,0) if screen dimensions cannot be retrieved.

## How to test it

**Steps to Reproduce:** 
- Build the Linux Library from the main branch.
- Run the binary in a standard Ubuntu terminal: `./bin/neutralino-linux_x64 --load-dir-res`
- observer the **Gdk-Critical**  warning in the terminal.

**Output of the Warning:**
```
Gdk-CRITICAL **: 14:28:17.432: gdk_monitor_get_workarea: assertion 'GDK_IS_MONITOR (monitor)' failed
```

**After the fix**
The assertion is no longer triggered.

## Next steps

None.

## Deploy notes

None.

### DIFF
--- a/api/window/window.cpp
+++ b/api/window/window.cpp
@@ -157,10 +157,14 @@ pair<int, int> __getCenterPos(bool useConfigSizes = false) {
         height = opt.height;
     }
     #if defined(__linux__) || defined(__FreeBSD__)
-    GdkRectangle screen;
-    gdk_monitor_get_workarea(gdk_display_get_primary_monitor(gdk_display_get_default()), &screen);
-    x = (screen.width - width) / 2;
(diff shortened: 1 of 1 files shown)


shalithasuranga (approved):
Thanks so much :tada:

PR #1644, opened 2026-03-16
fix: prevent memory leak and invalid state when createWatcher() fails

### LINKED ISSUE #1600: Returns invalid watch ID on failed addWatch
`Neutralino.filesystem.createWatcher` does not handle failures from the underlying watcher (e.g., non-existent or inaccessible paths). It still registers and returns an invalid watch ID, which can leave the watcher in an inconsistent state.

**To Reproduce**  
1. Create a Neutralinojs app and enable the native API.  
2. Call `Neutralino.filesystem.createWatcher("/path/does/not/exist")`.  
3. Observe that a watch ID is returned.  
4. Call `Neutralino.filesystem.removeWatcher(<returned_id>)` and see it fail or behave inconsistently.

**Expected behavior**  
`createWatcher` should return an error (or a failure result) when it cannot register the path, and it should not store an invalid watcher internally.

Not applicable.

**Specifications**  
- OS: macOS 14.x (Apple Silicon)  
- Neutralinojs version: [fill from `neu version`]  
- Neutralinojs client library version: [fill from `neu version`]  
- Neutralinojs CLI version: [fill from `neu version`]

**Additional context**  
This happens when the path does not exist or is not watchable. The returned ID can be invalid (e.g., `-1`), but it still gets stored and returned, which can lead to leaks and future API failures.

zebra0307:
If `efsw::addWatch()` returns a negative ID (invalid path, inaccessible path, etc.), the listener is now deleted and returned immediately without storing in `watchListeners`, preventing memory leaks and inconsistent watcher state.

Fixes: #1600

## Description
When `efsw::addWatch()` fails it returns a negative `WatchID`. Previously this negative ID and its allocated listener were unconditionally stored in `watchListeners`, causing memory leaks and leaving the watcher map in an inconsistent state where [removeWatcher()](cci:1://file:///d:/openSource/neutralinojs/api/fs/fs.cpp:267:0-276:1) would find the bad ID but fail to clean it up properly.

## Changes proposed

- Check `addWatch()` return value before registering the listener in `watchListeners`
- If negative, `delete listener` immediately and return the error code — nothing is stored

## How to test it

- Call `Neutralino.filesystem.createWatcher("/path/that/does/not/exist")`
- Observe that an error is returned (`NE_FS_UNLCWAT`) and no watcher ID is registered
- Call `Neutralino.filesystem.getWatchers()` and confirm the invalid path does not appear in the list

## Next steps

None.

## Deploy notes

None.

### DIFF
--- a/api/fs/fs.cpp
+++ b/api/fs/fs.cpp
@@ -256,6 +256,10 @@ long createWatcher(const string &path) {
 
     __WatcherListener* listener = new __WatcherListener();
     efsw::WatchID watcherId = fileWatcher->addWatch(path, listener, true);
+    if(watcherId < 0) {
+        delete listener;
+        return (long)watcherId;
+    }
     watchListeners[watcherId] = make_pair(listener, path);
     fileWatcher->watch();
     return (long)watcherId;


shalithasuranga:
Thanks for sending this. We need to fix #1600 differently. I was self-assigned to the issue to look it further..

PR #1647, opened 2026-03-17
Validate defaultPath directory before opening the save dialog

Swastik-Prakash1:
showSaveDialog fails silently on Windows when defaultPath points to a 
directory that does not exist. This adds a check using std::filesystem 
before calling pfd::save_file, and returns a NE_OS_INVDPATH error if 
the parent directory is missing.

To test: call Neutralino.os.showSaveDialog with a defaultPath whose 
parent folder does not exist and confirm an error is returned instead 
of silent failure.


shalithasuranga:
Looks like this belongs to the library. Thanks so much for sending this :)

PR #1651, opened 2026-03-18
fix: add mutex lock to updateSpawnedProcess to prevent use-after-free

### LINKED ISSUE #1622: os.updateSpawnedProcess is not thread-safe the missing mutex lock causes potential crash
**Describe the bug**

[updateSpawnedProcess()](cci:1://file:///Users/aadishah/Desktop/Neutralino/neutralinojs/api/os/os.cpp:217:0-238:1) in [api/os/os.cpp](cci:7://file:///Users/aadishah/Desktop/Neutralino/neutralinojs/api/os/os.cpp:0:0-0:0) reads from and dereferences pointers in the `spawnedProcesses` map without acquiring `spawnedProcessesLock`. Meanwhile, the cleanup thread spawned by [spawnProcess()](cci:1://file:///Users/aadishah/Desktop/Neutralino/neutralinojs/api/os/os.h:40:0-40:89) modifies the same map (`erase` + `delete`) under the lock. This creates a TOCTOU (time-of-check-time-of-use) race condition that can cause crashes or undefined behavior.

Specifically:
1. Thread A calls [updateSpawnedProcess](cci:1://file:///Users/aadishah/Desktop/Neutralino/neutralinojs/api/os/os.cpp:217:0-238:1), finds the process in the map, gets the raw pointer
2. Thread B (cleanup thread) acquires the lock, erases the entry, and `delete`s the pointer
3. Thread A dereferences the now-dangling pointer → **use-after-free / crash**

The same file correctly uses `lock_guard<mutex> guard(spawnedProcessesLock)` in [spawnProcess](cci:1://file:///Users/aadishah/Desktop/Neutralino/neutralinojs/api/os/os.h:40:0-40:89) (line 163) and [getSpawnedProcesses](cci:1://file:///Users/aadishah/Desktop/Neutralino/neutralinojs/api/os/os.cpp:394:0-407:1) controller (line 398), but [updateSpawnedProcess](cci:1://file:///Users/aadishah/Desktop/Neutralino/neutralinojs/api/os/os.cpp:217:0-238:1) is missing it entirely.

**To Reproduce**

1. Spawn a short-lived process via `Neutralino.os.spawnProcess` (e.g., `echo hello`)
2. Immediately call `Neutralino.os.updateSpawnedProcess` with a `stdIn` or `stdInEnd` event on the returned process ID
3. If the cleanup thread runs between the [find()](cci:1://file:///Users/aadishah/Desktop/Neutralino/neutralinojs/chrome.cpp:47:0-98:1) check and the pointer dereference, the app crashes

Note: This is timing-dependent and may not reproduce on every attempt, but is a real correctness issue under concurrent load.

**Expected behavior**

[updateSpawnedProcess](cci:1://file:///Users/aadishah/Desktop/Neutralino/neutralinojs/api/os/os.cpp:217:0-238:1) should acquire `spawnedProcessesLock` before accessing the `spawnedProcesses` map, ensuring the process entry and pointer remain valid for the duration of the operation.

Xavrir:
## Problem

`updateSpawnedProcess()` reads from the `spawnedProcesses` map and dereferences process pointers without holding `spawnedProcessesLock`. The cleanup thread spawned by `spawnProcess()` modifies the same map (`erase` + `delete`) under the lock, creating a TOCTOU race condition:

1. Thread A calls `updateSpawnedProcess`, finds the process, gets the raw pointer
2. Thread B (cleanup) acquires the lock, erases the entry, deletes the pointer
3. Thread A dereferences a dangling pointer

## Fix

Add `lock_guard<mutex> guard(spawnedProcessesLock)` at the top of `updateSpawnedProcess`, matching the existing pattern in `spawnProcess` (line 163) and `getSpawnedProcesses` (line 398).

The lock scope covers the full function body so the process entry and pointer remain valid for the duration of the operation. No deadlock risk since TinyProcessLib's `kill()`, `write()`, and `close_stdin()` don't acquire this lock internally.

Closes #1622

### DIFF
--- a/api/os/os.cpp
+++ b/api/os/os.cpp
@@ -216,6 +216,7 @@ pair<int, int> spawnProcess(string command, const os::ChildProcessOptions &optio
 }
 
 bool updateSpawnedProcess(const os::SpawnedProcessEvent &evt) {
+    lock_guard<mutex> guard(spawnedProcessesLock);
     if(spawnedProcesses.find(evt.id) == spawnedProcesses.end()) {
         return false;
     }


shalithasuranga (approved):
Thanks so much :tada:

PR #1652, opened 2026-03-19
feat: Added abillity to set darkMode on Windows

ZekoAR:
## Description
Added the abillity to set darkMode runtime and a config to force darkMode on Windows.

## Changes proposed
- Added window.darkMode config
- Added setDarkMode JS function
- Added --window-dark-mode

## How to test it
- set window.darkMode config to true and observe dark window frame on windows with light mode set.

- run setDarkMode(true/false) and observe window frame change

## Next steps
None.

## Deploy notes
None.


ZekoAR:
@shalithasuranga  Did you reject the PR?
The abillity to force dark mode is quite important if the app has the abillity to choose mode in config (or even a "dark mode only" app). Lot's of dev tools on Windows operates like this. For instance VSCode and Vistual Studio.

Without the functionallity in my PR, creating such options this is not possible for Neutralino apps on Windows.

Let me know what (if anything) can be done to the PR to be acceptable. :-)

PR #1658, opened 2026-03-22
Replace sleep interval 20000ms with named constant in __wait()

aarav2912:
**Description**
The __wait() function in main.cpp uses a hardcoded number 20000ms as the sleep interval. Random numbers reduce code readability and make future changes harder to track. This PR replaces it with a named constant defined alongside the other app-level macros at the top of the file.

**Changes proposed**
Added #define NEU_APP_WAIT_INTERVAL_MS 20000 near the top of main.cpp alongside existing macro definitions
Replaced the inline 20000ms literal in __wait() with chrono::milliseconds(NEU_APP_WAIT_INTERVAL_MS)

**How to test it**
Build the project and verify it compiles without warnings or errors
Run the app in browser or cloud mode (both trigger __wait()) and confirm the app continues to behave normally

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -26,6 +26,7 @@
 #define NEU_APP_LOG_FILE "/neutralinojs.log"
 #define NEU_APP_LOG_FORMAT "%level %datetime %msg %loc %user@%host"
 #define ELPP_THREAD_SAFE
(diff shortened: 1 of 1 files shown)


shalithasuranga (approved):
Thanks so much :tada:

PR #1660, opened 2026-03-22
Skip path replacement in applyPathConstants when os::getPath returns empty.

aarav2912:
**Title:**
```
Fix silent empty path replacement in applyPathConstants
```

**Description:**
```
## Description
In fs.cpp, applyPathConstants() calls os::getPath() to resolve 
path variables like ${NL_OSDATAPATH}. However, os::getPath() 
returns an empty string when it fails to resolve a path (e.g., 
on certain platforms or configurations). Previously, this empty 
string was still passed to regex_replace, silently replacing the 
path variable with nothing — resulting in broken file paths with 
no error thrown.

This PR adds a guard to skip the replacement entirely when 
os::getPath() returns an empty string, preserving the original 
path variable in the string instead of corrupting it silently.

## Changes proposed
- Extracted os::getPath() result into a local variable resolvedPath
- Added an empty check before performing regex_replace in applyPathConstants

## How to test it
- Use a path variable like ${NL_OSDATAPATH} in neutralino.config.json
- Verify it resolves correctly on supported platforms
- On a platform where the path cannot be resolved, verify the 
  original variable is preserved rather than replaced with an empty string

## Next steps
None.

## Deploy notes
None.

### DIFF
--- a/api/fs/fs.cpp
+++ b/api/fs/fs.cpp
@@ -355,13 +355,16 @@ fs::DirReaderResult readDirectory(const string &path, bool recursive) {
 string applyPathConstants(const string &path) {
     string newPath = regex_replace(path, regex("\\$\\{NL_PATH\\}"), settings::getAppPath());
 
-    vector<string> pathNames = {"data", "cache", "documents", 
+    vector<string> pathNames = {"data", "cache", "documents",
                     "pictures", "music", "video", "downloads",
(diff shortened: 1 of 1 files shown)


prasu-baran:
hi !
here we can think of giving some warning kind of thing that would make failures even more visible.

PR #1661, opened 2026-03-22
Fix missing mutex lock in updateSpawnedProcess

aarav2912:
**Title:**
```
Fix missing mutex lock in updateSpawnedProcess
```

**Description:**
```
## Description
In os.cpp, updateSpawnedProcess() reads and accesses the 
spawnedProcesses map without holding spawnedProcessesLock. 
Meanwhile, spawnProcess() runs a detached thread that erases 
entries from spawnedProcesses when a process exits. This creates 
a race condition — the process entry could be erased between the 
find() check on line 219 and the map access on line 223, leading 
to undefined behavior.

All other functions that access spawnedProcesses (spawnProcess, 
getSpawnedProcesses, and the detached thread) correctly acquire 
spawnedProcessesLock before accessing the map. This PR makes 
updateSpawnedProcess consistent with that pattern.

## Changes proposed
- Added lock_guard<mutex> guard(spawnedProcessesLock) at the 
  start of updateSpawnedProcess, consistent with all other 
  functions that access spawnedProcesses

## How to test it
- Spawn multiple processes simultaneously
- Send events to update/kill spawned processes while others 
  are exiting concurrently
- Verify no crashes or undefined behavior occur

## Next steps
None.

## Deploy notes
None.
```

### DIFF
--- a/api/os/os.cpp
+++ b/api/os/os.cpp
@@ -126,7 +126,7 @@ os::CommandResult execCommand(string command, const os::ChildProcessOptions &opt
         commandResult.stdErr += string(bytes, n);
     };
 
-    if(!options.background && options.envs.empty()) { 
+    if(!options.background && options.envs.empty()) {
         childProcess = new TinyProcessLib::Process(CONVSTR(command), CONVSTR(options.cwd), stdOutHandler, stdErrHandler, !options.stdIn.empty());
     }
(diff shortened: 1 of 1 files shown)


shalithasuranga:
Closing since another PR already fixed this issue. Thanks so much :)

PR #1666, opened 2026-03-25
feat: add "home" and "desktop" paths to os.getPath()

### LINKED ISSUE #1568: os.getPath() has no support for home and desktop paths

achill06:
## Description
Closes #1568

os.getPath() was missing two commonly needed paths. "desktop" was already available in the bundled sago library, but never wired up. "home" required a new private helper since sago does not expose it.

## Changes proposed
- Added "desktop" path using sago::getDesktopFolder()
- Added `home` path resolution:
  - Uses `$HOME` or `getpwuid()` on Unix systems
  - Uses `SHGetKnownFolderPath` on Windows
- Added null checks for `getpwuid()` to prevent potential crashes
- Added pwd.h header for Linux/macOS and shlobj.h for Windows

## How to test it
Run the following in any Neutralino app's dev console:

```
(async () => {
  console.log("home:", await Neutralino.os.getPath("home"));
  console.log("desktop:", await Neutralino.os.getPath("desktop"));
})();
```

Expected: correct home and desktop paths for the current user

## Next steps

None.

## Deploy notes

None.

### DIFF
--- a/api/os/os.cpp
+++ b/api/os/os.cpp
@@ -241,7 +241,7 @@ bool updateSpawnedProcess(const os::SpawnedProcessEvent &evt) {
 }
 
 static string __getHomePath() {
-    #if defined(__linux__) || defined(__APPLE__)
+    #if defined(__linux__) || defined(__APPLE__) || defined(__FreeBSD__)
         const char* home = getenv("HOME");
         if(getuid() != 0 && home)
             return string(home);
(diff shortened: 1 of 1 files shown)


shalithasuranga:
Thanks for sending this. I had to close this since the same feature was added by a previous pull request :)

PR #1672, opened 2026-03-27
feat(window): redirect external URLs to system browser

### LINKED ISSUE #1555: Handle window.open / NewWindowRequested by opening URL in system browser

achill06:
## Description
Closes #1555

When a Neutralinojs app attempts to open an external link (via `window.open(url)` or `<a target="_blank">`), the webview has no handler for the new-window request. It is silently dropped on Windows, and on Linux/macOS, the behavior is undefined or trapped inside an internal webview.

## Changes proposed
- **lib/webview/webview.h:** Added a `newWindowHandler` callback and updated WebView2, WebKitGTK, and WKWebView backends to intercept "new window" requests and block the internal popup.
 - **api/window/window.cpp:** Attached the new handler to delegate the intercepted external URLs to `os::open()`. Internal URLs (`localhost` and `127.0.0.1`) are explicitly excluded to preserve app-internal navigation.

## How to test it
- Add a standard HTML link to a test app: 
```
<a href="https://example.com" target="_blank">Test HTML Link</a>
```
 - Add a JS button to a test app: 
 ```
<button onclick="window.open('https://example.com')">Test JS Link</button>
```
Expected: 
 - On click, it should successfully open in the OS's default web browser (e.g., Chrome, Firefox) rather than a secondary Neutralinojs window.
 - Standard internal navigation (localhost URLs) should work as expected within the app.

## Next steps

None.

## Deploy notes

None.

### DIFF
--- a/api/window/window.cpp
+++ b/api/window/window.cpp
@@ -54,6 +54,7 @@
 #include "api/fs/fs.h"
 #include "api/debug/debug.h"
 #include "api/computer/computer.h"
+#include "api/os/os.h"
 
 using namespace std;
 using json = nlohmann::json;
@@ -634,6 +635,10 @@ bool __createWindow() {
 );
 
     nativeWindow->setEventHandler(&window::handlers::windowStateChange);
+
+    nativeWindow->setNewWindowHandler([](const std::string& url) {
+        os::open(url);
+    });
 
     if(windowProps.injectGlobals) 
         nativeWindow->init(settings::getGlobalVars() + "var NL_GINJECTED = true;");
--- a/lib/webview/webview.h
+++ b/lib/webview/webview.h
@@ -320,7 +320,7 @@ public:
             const char* uri = webkit_uri_request_get_uri(request);
             if(uri) {
                 std::string uriStr(uri);
-                if(uriStr.find("http://localhost") != 0 && newWindowHandler) {
+                if(uriStr.find("http://localhost") != 0 && uriStr.find("http://127.0.0.1") != 0 &&newWindowHandler) {
                     newWindowHandler(uriStr);
                 }
             }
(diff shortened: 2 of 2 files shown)


shalithasuranga (approved):
Thanks so much :tada:

Issue #1163, opened 2023-10-18
NSWindow warning: adding an unknown subview


gzl10g:
**Describe the bug**
When run with "neu run --frontend-lib-dev ". I get this error, and the logs stop viewing in console.

neutralino-mac_arm64[82786:3790932] NSWindow warning: adding an unknown subview: <WKInspectorWKWebView: 0x13e74ff80>. Break on NSLog to debug.

*Screenshots/Screen recordings**
 neutralino-mac_arm64[82786:3790932] NSWindow warning: adding an unknown subview: <WKInspectorWKWebView: 0x13e74ff80>. Break on NSLog to debug.
[nl]  neutralino-mac_arm64[82786:3790932] Call stack:
[nl] (
[nl]    0   AppKit                              0x000000019bbd1790 -[NSThemeFrame addSubview:] + 116
[nl]    1   AppKit                              0x000000019bbd1470 -[NSView addSubview:positioned:relativeTo:] + 176
[nl]    2   AppKit                              0x000000019bbd133c -[NSThemeFrame addSubview:positioned:relativeTo:] + 52
[nl]    3   WebKit                              0x00000001b9a06d10 _ZN6WebKit19WebInspectorUIProxy14platformAttachEv + 340
[nl]    4   WebKit                              0x00000001b9be5874 _ZN6WebKit19WebInspectorUIProxy4openEv + 156
[nl]    5   WebKit                              0x00000001b9e49948 _ZN6WebKit19WebInspectorUIProxy17didReceiveMessageERN3IPC10ConnectionERNS1_7DecoderE + 184
[nl]    6   WebKit                              0x00000001b9f42210 _ZN3IPC18MessageReceiverMap15dispatchMessageERNS_10ConnectionERNS_7DecoderE + 264
[nl]    7   WebKit                              0x00000001b9b8afec _ZN6WebKit15WebProcessProxy17didReceiveMessageERN3IPC10ConnectionERNS1_7DecoderE + 40
[nl]    8   WebKit                              0x00000001b9f3da00 _ZN3IPC10Connection15dispatchMessageENSt3__110unique_ptrINS_7DecoderENS1_14default_deleteIS3_EEEE + 332
[nl]    9   WebKit                              0x00000001b9f3deac _ZN3IPC10Connection24dispatchIncomingMessagesEv + 284
[nl]    10  JavaScriptCore                      0x00000001b2da40c4 _ZN3WTF7RunLoop11performWorkEv + 200
[...]

PR #1691, opened 2026-03-29
fix(macOS): resolve NSWindow unknown subview warning when inspector is enabled (#1163)

itssagarK:
This PR fixes a macOS issue where enabling the Web Inspector would trigger the warning:

NSWindow warning: adding an unknown subview: WKInspectorWKWebView

After this warning, console logs would also stop appearing.

---

### Root Cause
On macOS 13.3+, WKWebView needs to be explicitly marked as inspectable for the Web Inspector to work correctly.

In the previous implementation:
- `developerExtrasEnabled` was enabled
- But WKWebView itself wasn’t marked as inspectable
- As a result, WebKit tried to attach the inspector (`WKInspectorWKWebView`) directly to the main NSWindow instead of opening it properly

---

### Fix
- Enabled `_setInspectable` using the Objective-C runtime (with `respondsToSelector` checks for safety)
- Made sure the inspector only opens when `openInspector` is explicitly set to true
- Prevented the inspector from attaching automatically to the main window
- Updated the default behavior by setting `openInspectorOnStartup = false`

---

### Design Notes
- Uses private WebKit APIs (`_setInspectable`, `_inspector`) since there is no public alternative
- All calls are guarded to avoid issues on unsupported macOS versions
- Change is limited to macOS (`WEBVIEW_COCOA`) and does not affect other platforms
- Keeps behavior consistent with GTK and Windows, where the inspector opens only when requested

---

### Testing
This is a macOS-specific change, so I wasn’t able to test it locally.

Expected behavior:
- With `enableInspector: true` and `openInspectorOnStartup: false`  
  → No warning, app runs normally, inspector does not open automatically  

- With `enableInspector: true` and `openInspectorOnStartup: true`  
  → No warning, inspector opens in a separate window  

It would be helpful if this could be verified on macOS (Ventura/Sonoma).

---

### Risks
- Relies on private WebKit APIs, which could change in future macOS versions

Mitigation:
- Uses `respondsToSelector` checks to ensure safe usage
- Limited to inspector/debug-related functionality only


shalithasuranga (changes requested):
Please send only the related code change. Thank you :)

PR #1693, opened 2026-03-30
fix(macOS #1692 ): prevent crash in multi-window scenarios caused by duplicate Objective-C class allocation 

### LINKED ISSUE #1692: macOS crash when creating multiple windows due to objc_allocateClassPair returning NULL
I found a potential crash issue in the macOS WebView implementation when multiple windows are created.

In lib/webview/webview.h, objc_allocateClassPair is called with fixed class names like "AppDelegate" and "WindowDelegate". This works for the first window, but when a second window is created (e.g., using window.create()), objc_allocateClassPair returns NULL because a class with the same name is already registered in the Objective-C runtime.

The code then proceeds to use this NULL pointer (e.g., calling class_addMethod), which can lead to a crash or undefined behavior.

This issue can occur in:
- multi-window applications
- webview reinitialization
- any scenario where the same class is allocated more than once

---

### Why this happens
objc_allocateClassPair does not allow creating a class with an existing name. When called again with the same name, it returns NULL.

The current implementation does not check for this condition before using the returned value.

---

### Suggested fix
Before calling objc_allocateClassPair, check if the class already exists using objc_getClass:

```cpp
Class cls = (Class)objc_getClass("AppDelegate");
if (!cls) {
    cls = objc_allocateClassPair((Class)"NSResponder"_cls, "AppDelegate", 0);
    // add methods and register class
}

itssagarK:
This PR fixes a macOS #1692 crash that occurs when multiple windows are created or when the WebView is initialized more than once.

In the current implementation, objc_allocateClassPair is used to create Objective-C classes such as AppDelegate and WindowDelegate. However, these classes are created using fixed names.

In the Objective-C runtime, class names must be unique. When objc_allocateClassPair is called again with the same class name, it returns NULL because the class is already registered.

The existing code does not check for this case and continues execution using the returned value. This leads to situations where methods are added to a NULL class pointer, which can cause crashes or undefined behavior.

This issue is more likely to appear in:
- multi-window applications (e.g., using window.create)
- repeated WebView initialization
- any scenario where the same delegate classes are created more than once

### Root Cause

- objc_allocateClassPair does not allow duplicate class names
- repeated calls with the same name return NULL
- NULL is not handled before calling class_addMethod and related APIs

### Solution

This PR updates the logic to safely handle class creation:

- Before allocating a class, it checks if the class already exists using objc_getClass
- If the class exists, it is reused instead of being created again
- If the class does not exist, it is allocated, methods are added, and it is registered
- Method and protocol registration is performed only once during initial creation

### Impact

- Prevents crashes caused by NULL class pointers
- Ensures stable behavior in multi-window scenarios
- Improves reliability of macOS WebView lifecycle handling
- Aligns implementation with Objective-C runtime requirements

---
This fix ensures that Objective-C classes are handled correctly across multiple window instances and prevents runtime crashes related to duplicate class allocation.
---
Linked Issue
Fixes #1692


shalithasuranga (changes requested):
Please send only the related code change. Thank you :)

PR #1701, opened 2026-04-01
fix: prevent buffer underflow and overflow in os.setTray (fixes #1698)

itssagarK:
This PR fixes a memory safety issue in the `os.setTray` implementation where missing bounds validation can lead to buffer underflow and overflow errors.

---

### Problem

The current implementation doesn't validate the `menuItems` array size before processing it, creating two critical issues:

#### 1. Buffer Underflow (Empty Array)

When `menuItems` is empty:
- `menuCount` becomes 0
- The code accesses `menus[menuCount - 1]`, which evaluates to `menus[-1]`
- This causes invalid memory access and application crashes

#### 2. Buffer Overflow (Too Many Items)

When `menuItems` exceeds the limit:
- The `menus` array has a fixed size of `NEU_MAX_TRAY_MENU_ITEMS` (50)
- Passing more than 50 items causes the loop to write beyond array bounds
- This corrupts memory and leads to undefined behavior

---

### Solution

This PR adds proper bounds validation to handle both cases safely:

**Changes made:**

```cpp
int menuCount = input["menuItems"].size();

if (menuCount == 0) {
    return output;
}

if (menuCount > NEU_MAX_TRAY_MENU_ITEMS) {
    menuCount = NEU_MAX_TRAY_MENU_ITEMS;
}

int i = 0;
for (const auto &menuItem : input["menuItems"]) {
    if (i >= menuCount) break;
    // populate menus
    i++;
}
```

**Key improvements:**
- Prevents invalid access when menuItems is empty
- Limits item count to avoid overflow
- Ensures loop respects bounded size
- Keeps existing behavior unchanged for valid inputs

---

### Testing

The fix was tested with a sample Neutralino application using the updated binary.

#### Test Case 1: Empty Menu

```js
await Neutralino.os.setTray({ menuItems: [] });
```

**Result:** ✅ No crash, handled safely without any issues

#### Test Case 2: Oversized Menu

```js
await Neutralino.os.setTray({
  menuItems: new Array(60).fill({ text: "item" })
});
```

**Result:** ✅ No crash, excess items ignored safely, application remains stable

Both test cases executed successfully without errors or memory issues.

---

### Impact

- **Security:** Prevents potential memory corruption vulnerabilities
- **Stability:** Eliminates crashes from invalid array access
- **Usability:** Applications can safely pass any menu size without worrying about crashes
- **Backward Compatibility:** Normal use cases are unaffected

---

### Linked Issue

Fixes #1698


shalithasuranga (changes requested):

PR #1703, opened 2026-04-01
fix: prevent memory leak in fs.createWatcher on failure (fixes #1700)

itssagarK:
This PR fixes a memory leak in `fs::createWatcher` that occurs when directory watch creation fails. The function did not free allocated resources or check the result of the operation.

---

### Problem

The current implementation allocates a listener object before calling `addWatch`:

```cpp
__WatcherListener* listener = new __WatcherListener();
efsw::WatchID watcherId = fileWatcher->addWatch(path, listener, true);
watchListeners[watcherId] = make_pair(listener, path);
```

When `addWatch` fails (invalid path, permission denied, etc.), it returns a negative error code. However, the code ignores this and continues execution, causing:

1. **Memory leak:** The allocated `listener` object is never deleted
2. **Map corruption:** A negative `watcherId` is stored in `watchListeners` map

If a user attempts to watch 100 invalid paths, 100 listener objects leak into memory with no way to recover them.

---

### Solution

Added a simple error check after `addWatch` is called:

```cpp
__WatcherListener* listener = new __WatcherListener();
efsw::WatchID watcherId = fileWatcher->addWatch(path, listener, true);

if (watcherId < 0) {
    delete listener;
    return (long)watcherId;
}

watchListeners[watcherId] = make_pair(listener, path);
```

This ensures:
- Memory is properly freed if watcher creation fails
- Invalid watcher IDs are never added to the map
- The error is returned to the caller immediately

---

### Testing

Tested with both valid and invalid directory paths.

#### Test Case 1: Invalid Path

```js
await Neutralino.filesystem.createWatcher("/invalid/nonexistent/path");
```

 **Result:**
- Function returns error code
- No crash
- No memory leaks
- No corrupt entries in map

#### Test Case 2: Valid Path

```js
await Neutralino.filesystem.createWatcher("./");
```

 **Result:**
- Watcher created successfully
- File change events received correctly
- No performance impact

---

### Impact

- **Fixes memory leaks** when watcher creation fails
- **Prevents map corruption** from negative IDs
- **Improves error handling** with proper resource cleanup
- **Enhances stability** of filesystem watcher API

---

### Linked Issue

Fixes #1700


shalithasuranga (changes requested):

PR #1717, opened 2026-04-05
Fix race condition in spawnedProcesses map access

PratikshaJ-16:
## Description

This PR fixes a potential race condition in the process management logic involving the shared `spawnedProcesses` map.
Previously, `updateSpawnedProcess` accessed the map without synchronization and used direct indexing, which could lead to unsafe concurrent access and unintended insertions.

## Changes proposed

* Added mutex protection (`spawnedProcessesLock`) using `std::lock_guard` to ensure thread-safe access
* Replaced direct map access (`spawnedProcesses[evt.id]`) with iterator-based lookup (`find`)
* Accessed the process using `it->second` instead of direct indexing
* Added a null check for `childProcess` before usage

## How to test it

* Spawn multiple processes
* Trigger events like `exit` and `stdin`
* Verify:

  * No crashes or undefined behavior occur
  * Process handling works correctly

## Next steps

* Review similar shared data structures for proper synchronization

### DIFF
--- a/api/os/os.cpp
+++ b/api/os/os.cpp
@@ -216,11 +216,18 @@ pair<int, int> spawnProcess(string command, const os::ChildProcessOptions &optio
 }
 
 bool updateSpawnedProcess(const os::SpawnedProcessEvent &evt) {
-    if(spawnedProcesses.find(evt.id) == spawnedProcesses.end()) {
-        return false;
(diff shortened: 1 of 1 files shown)


shalithasuranga:
Closing since the same fix was added by a previous pull request. Thanks for sending this :)

PR #1725, opened 2026-04-12
Improve getEnv handling and explore command validation for execCommand

itssagarK:
This PR implements critical security hardening and stability improvements for the os.cpp module. The changes focus on preventing command injection, improving environment variable retrieval, and modernizing memory management using RAII principles.

### 1. Command Injection Mitigation (Critical)
- **Security Helpers:** Added `containsShellMetachar` to detect dangerous characters (like `&`, `;`, `|`, etc.) and `escapeWindowsCommand` to handle Windows-specific shell escaping.
- **Input Validation:** Updated the `execCommand` controller to validate strings before execution. Suspicious inputs now trigger a specific `NE_OS_INVINPT` error.
- **Error Registration:** Updated `errors.h` and `errors.cpp` to include the new `NE_OS_INVINPT` error code and user-friendly message.

### 2. Robust Environment Variable Retrieval (Improvement)
- **Windows API Refactor:** Updated `getEnv` to use `GetEnvironmentVariable` with explicit error checking via the `DWORD` return value.
- **Graceful Failures:** The function now correctly returns an empty string if a variable is missing or if the system call fails, improving upon the previous ternary implementation.

### 3. Memory Management & Stability (Medium)
- **Smart Pointers:** Replaced raw `TinyProcessLib::Process` pointers with `std::unique_ptr` in the `execCommand` logic.
- **Automatic Cleanup:** This migration ensures process resources are automatically reclaimed when the function scope ends, eliminating potential memory leaks.

##  How to Test
1. **Security Verification:** Attempted to run a command containing metacharacters: `Neutralino.os.execCommand({ command: 'echo "test" & whoami' })`. The command was successfully intercepted.
2. **Environment Verification:** Successfully retrieved the system PATH using `Neutralino.os.getEnv("PATH")` as seen in the console logs.

image

image


prasu-baran:
Hi ! ,
many commands are broken by this method and by doing this we can break many legit usage 
`echo "hello world"` will  also be blocked via this command  due to presence of this code
 const string dangerousChars = "\"&;|$`<>|(){}[]\\^~!#%*?";
 I think this is just making over complicating the already existing system and it is not backend exposed to internet or something . So right now it's need is first to be discussed with maintainers first then should implement.

itssagarK:
Hi  @prasu-baran
 
Thanks for the feedback , I get your point.

I added this mainly thinking about cases where user input might be passed into `execCommand` without proper checks, so I tried adding a safety layer at the framework level.
But I agree this approach is too strict and breaks valid commands, which isn’t ideal here. I see that this is more of a design decision than a direct fix.

I’ll rethink the approach and decide the next step.

shalithasuranga:
At this stage, we are not planning to implement command validation and filtering. The developer is responsible for blocking command APIs whenever needed using permissions configuration. I'll close this for now. Let's re-open this if we decide to implement command validation :) Thanks so much for sending this :)

PR #1740, opened 2026-05-17
(Fix): optional title handling for OS dialogs

### LINKED ISSUE #1727: Neutralino.os.showOpenDialog expects title even tho it is optional as the Docs say
**Describe the bug**

**To Reproduce**
1. Createa a simple app with the following function (can be via button press or anything)
```ts
await Neutralino.os.showOpenDialog(
            {filters: [
                {name: "Name", extensions: ["exe"]}
            ]}
        )
```
2. See error of telling you that a parameter is missing, which is the title-parameter

**Expected behavior**
As the parameter is optional according to the docs "title String (optional): Title of the dialog.", the error shouldn't happen

Screenshots can be added later. Found the bug with a minimum of time left for me.

**Specifications**  
 - Windows 11 x64
 -  Binaries and Lib v6.7.0
 -  CLI v11.7.1

**Additional context**
Not sure if there is any context for that. Trying to open a dialogue without giving it an "optional" titel lead to an exception on my end.

Vinayak9769:
## Description

  Fixes optional title handling for OS file dialogs when dialog options are
  provided without a title.

Closes #1727

  ## Changes proposed

  - Normalize dialog input before reading title, filters, defaultPath, and
    related options.
  - Preserve existing default titles for open, save, and folder dialogs when
    title is omitted.

  ## How to test it

  - Build the project with cmake --build build.
  - Call Neutralino.os.showOpenDialog({ filters: [{ name: "Name", extensions:
    ["exe"] }] }) and verify it no longer reports a missing title parameter.

  ## Next steps

  None.

  ## Deploy notes

  None.

### DIFF
--- a/api/os/os.cpp
+++ b/api/os/os.cpp
@@ -329,6 +329,20 @@ vector<string> __extensionsToVector(const json &filters) {
     return filtersV;
 }
 
+json __normalizeDialogInput(const json &input) {
+    if(!helpers::hasField(input, "title") || !input["title"].is_object()) {
+        return input;
+    }
+
+    json normalizedInput = input["title"];
+    for(auto &entry: input.items()) {
+        if(entry.key() != "title") {
+            normalizedInput[entry.key()] = entry.value();
+        }
+    }
+    return normalizedInput;
+}
+
 json execCommand(const json &input) {
     json output;
     if(!helpers::hasRequiredFields(input, {"command"})) {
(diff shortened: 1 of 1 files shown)


Vinayak9769:
@shalithasuranga Hi! Could I get a review on this PR when you have a chance? Thanks! :)

shalithasuranga:
Hello, thanks for sending this. The first parameter is optional if we just call the function, but if we need to use the second one, then we should pass `undefined` to the first:
```js
showOpenDialog();
showOpenDialog(undefined, {options})
```

This is an issue with the function definition. Ideally, we should use `showOpenDialog(options)` since the titile can be optional. But it will trigger a breaking release. I think we can check this later. Thanks

PR #1741, opened 2026-05-19
Get gpu info

prasu-baran:
## Description

Implemented cross-platform computer.getGPUInfo() API using native platform-specific GPU information providers.
## Changes proposed

 - Added GPU information retrieval using DXGI, IOKit, and Linux DRM interfaces.
 - Exposed computer.getGPUInfo through Neutralino router, permissions, and JavaScript SDK.
 
## How to test it

Build Neutralino core, run application, verify returned GPU information object values.

## Next steps

None.

## Deploy notes

None.


shalithasuranga:
This is still a draft, commenting to save your time :). Please fetch GPU info from https://github.com/neutralinojs/neutralinojs/tree/main/lib/infoware (we need to update the library files first seems). Note that we can try using existing libraries first before deciding to implement things ourselves ;) Thanks

lib/infoware/src/gpu/blank_all.cpp
```diff
@@ -17,9 +17,358 @@
...
+#ifdef __APPLE__
+#include <CoreFoundation/CoreFoundation.h>
+#include <IOKit/IOKitLib.h>
+#endif
+
+using namespace std;
+
+
+#ifdef __linux__
+static vector<string> drmCardDevicePaths() {
```
shalithasuranga: Seems like you are modifying the original library with your own implementation. The idea is that we don't like to maintain device info logic if libraries are available. Please research getting GPU info from the original library. If the current library has issues on a specific platform, you can use https://github.com/lfreist/hwinfo

Please let me know if you have questions

Thanks :tada:

shalithasuranga (changes requested):

prasu-baran:
okay got to your point now i think it aligns as stated earlier . Guide me if going in wrong direction

lib/infoware/src/gpu/d3d.cpp
```diff
@@ -16,11 +16,29 @@
 #include "infoware/detail/memory.hpp"
 #include "infoware/detail/winstring.hpp"
 #include "infoware/gpu.hpp"
-#include "infoware/pci.hpp"
 #include <d3d11.h>
 #include <memory>
 
 
+static iware::gpu::vendor_t vendor_from_id(const UINT id) {
```
shalithasuranga: Hello, this file is not the same as: https://github.com/ThePhD/infoware/blob/main/src/gpu/d3d.cpp. Could you let me know why we can use the original library? Thanks

shalithasuranga (changes requested):

prasu-baran:
I changed the vendored d3d.cpp because file depends on iware::pci::identify_device(), but Neutralino’s vendored Infoware copy doesn’t include the PCI implementation/source, so enabling INFOWARE_USE_D3D would fail at link time.
I realised it will be  better not to patch third-party code . 
I am new to it that's why taking little more time . Looking to correct it

shalithasuranga:
Thanks for your reply. Then, this will only work on Windows (via D3D). Could you check the possibility of using hwinfo?

shalithasuranga (approved):
Thanks so much for updating based on requests :tada:

PR #1752, opened 2026-05-27
fix: guard stoi calls to prevent crashes on invalid inputs (#1583)

### LINKED ISSUE #1583: Fix unguarded `stoi()` calls to prevent crashes on invalid inputs

JhaSourav07:
## Description

Unguarded `std::stoi()` calls in `settings.cpp` and `resources.cpp` caused the Neutralinojs process to terminate immediately when given non-numeric input — for example, passing `--port=notanumber` on the command line or loading a malformed resource bundle. This fix replaces those bare calls with safe, status-code-based error handling that matches the project's no-exceptions coding style.

Fixes #1583 

## Changes proposed

 - Added `helpers::strToInt()` — a reusable safe integer parse helper returning `bool` (exception never escapes to callers)
 - Added error code `NE_CF_INVINTVAL` for invalid integer config override values
 - Guarded the `stoi()` call in `settings::init()` — invalid integer CLI overrides (e.g. `--port=abc`, `--window-width=overflow`) are now logged and skipped; all other valid overrides in the same run still apply
 - Guarded both `stoi()` calls in `resources::__getFileFromBundle()` and `resources::__getFileFromEmbedded()` — malformed bundle offsets now return `NE_RS_TREEGER` instead of crashing (uses `strtoul` to also correctly handle offsets exceeding `INT_MAX`)

## How to test it

 - Run the config spec:
   ```bash
   cd spec && node index.js config
   ```
 - Manually pass an invalid integer CLI arg and confirm the app starts without crashing:
   ```bash
   ./bin/neutralino-linux_x64 --load-dir-res --port=notanumber
   ```
   Expected: app starts, `NE_CF_INVINTVAL` is logged, port falls back to the config default.

## Next steps

None.

## Deploy notes

None.

---

### Before:
image

### After:

image

### DIFF
--- a/errors.cpp
+++ b/errors.cpp
@@ -75,6 +75,7 @@ string __getStatusCodeString(const errors::StatusCode code) {
         case errors::NE_CF_UNBPRCF: return "NE_CF_UNBPRCF";
         case errors::NE_CF_UNSUPMD: return "NE_CF_UNSUPMD";
         case errors::NE_CF_UNBLWCF: return "NE_CF_UNBLWCF";
+        case errors::NE_CF_INVINTVAL: return "NE_CF_INVINTVAL";
     }
     return "NE_ST_NOTOK";
 }
@@ -145,6 +146,7 @@ string __findStatusCodeDesc(errors::StatusCode code) {
         case errors::NE_CF_UNBPRCF: return "Unable to parse the config file: %1";
         case errors::NE_CF_UNSUPMD: return "Unsupported mode: %1. The default mode (window) is selected.";
         case errors::NE_CF_UNBLWCF: return "Unable to load the window config file: %1";
+        case errors::NE_CF_INVINTVAL: return "Invalid integer value for config key '%1'. Skipping override.";
     }
     return "";
 }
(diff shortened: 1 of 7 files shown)


JhaSourav07:
@shalithasuranga hey, can you approve the workflow?

shalithasuranga:
Thanks so much for your work. I have to close this since we need to discuss this and implement a global solution for parsing correct JSON types. If we decide to use this approach, we'll reopen this pull request. Thanks :tada:

PR #1753, opened 2026-05-28
get UUID

prasu-baran:
## Description

Implemented cross-platform computer.getUUID() API using infoware system-level machine identifier support.

## Changes proposed

 - Added machine_uuid() implementation for Windows, Linux, and macOS platforms.
 - Exposed computer.getUUID() API through Neutralino native and JavaScript layers.
 
## How to test it

 - Build NeutralinoJS and execute Neutralino.computer.getUUID()

## Next steps

None.

## Deploy notes

None.


shalithasuranga (changes requested):
Please move the implementation from infoware (we are not planning to modify it with features, as I explained in another pull request) to `computer.cpp`. I think you can create a function called `__getMachineId`

Thanks, Let me know if you have questions :tada:

shalithasuranga (approved):
Thanks so much for updating as requested :)

PR #1754, opened 2026-05-28
 opacity in window

prasu-baran:
## Description

Adds runtime window opacity control across supported desktop platforms.

## Changes proposed

 - Implement window.setOpacity() native API with validation and tests.
## How to test it

 - Run specs/tests

## Next steps

None.

## Deploy notes

None.


shalithasuranga (changes requested):
Thanks for sending this. Please check why the test cases are failing :)

prasu-baran:
I think it is getting built from neutralino.js repo which is not having setOpacity API right now that might be the reason for the  failing testcases

PR #1772, opened 2026-06-04
get DiskInfo

prasu-baran:
## Description

Add computer.getDiskInfo() to report current disk usage information.
## Changes proposed

 - Add cross-platform disk information support through the computer API.
 - Implement Windows, Linux, and macOS disk usage retrieval logic.

## How to test it

 - Run specs/tests

## Next steps

None.

## Deploy notes

None.


lib/infoware/src/system/disk_info/disk_info_non_windows.cpp
```diff
@@ -0,0 +1,138 @@
...
+		while(true) {
+			if(getcwd(buffer.data(), buffer.size()))
+				return buffer.data();
+			if(errno != ERANGE)
+				return {};
+			buffer.resize(buffer.size() * 2);
+		}
+	}
+
+	static bool is_path_within_mount(const string& path, const string& mount_point) {
```
shalithasuranga: Thanks for sending this. Seems like you are trying to modify the infowave library. I think you misunderstood my feedback regarding this. If infowave doesn't support disk info, we can try another library without updating it with our own logic.

Please use https://github.com/lfreist/hwinfo to fetch disk info. 

Let me know if you have questions :tada:

shalithasuranga (changes requested):

prasu-baran:
Hi @shalithasuranga ,
I have made changes as suggested 
Kindly review
Thanks !

CMakeLists.txt
```diff
@@ -23,6 +23,52 @@ execute_process(
...
+set(HWINFO_NETWORK OFF CACHE BOOL "Enable hwinfo network information" FORCE)
+set(HWINFO_SOURCE_DIR "${PROJECT_SOURCE_DIR}/lib/hwinfo")
+if(NOT EXISTS "${HWINFO_SOURCE_DIR}/CMakeLists.txt")
+  include(FetchContent)
+  FetchContent_Declare(
+    lfreist_hwinfo
+    GIT_REPOSITORY https://github.com/lfreist/hwinfo.git
+    GIT_TAG main
+  )
+  FetchContent_GetProperties(lfreist_hwinfo)
```
shalithasuranga: Thanks so much for your updates. A small thing to do before merging this. Note that we include all libraries in `lib` and don't use remote repos for several reasons. You can now include `hwinfo/hwinfo.h` since I already added it to `lib`.

Please remove these changes and keep the original CMake file. Also, please include vendor and model as well (Check [this example](https://github.com/lfreist/hwinfo/blob/38e3816044e8490be5002356c958b958e848c26f/examples/system_infoMain.cpp#L117))

Thanks :tada:

shalithasuranga (changes requested):

shalithasuranga (approved):
Thanks so much for updating as requested :)

PR #1776, opened 2026-06-14
processID shown

prasu-baran:
## Description

Expose the app process ID through Neutralino.app.getProcessId().
## Changes proposed

 - Added Native API & test

## How to test it

 - Run specs/tests

## Next steps

None.

## Deploy notes

None.


shalithasuranga (approved):
Thanks for sending this. Looks great. I've updated your code to directly return the PID without the `pid` key in the object :)

PR #1784, opened 2026-06-25
implements openWith

prasu-baran:
## Description

Adds os.openWith to open files using a chosen application.
## Changes proposed

 - Added native openWith implementations 

## How to test it

 - Run specs/tests

## Next steps

None.

## Deploy notes

None.

### DIFF
--- a/api/os/os.cpp
+++ b/api/os/os.cpp
@@ -125,6 +125,71 @@ void open(const string &url) {
     #endif
 }
 
+bool openWith(const string &filePath, const string &appPath) {
+    #if defined(__linux__) || defined(__FreeBSD__)
+    vector<TinyProcessLib::Process::string_type> args = {
+        CONVSTR(appPath),
+        CONVSTR(filePath)
+    };
+    TinyProcessLib::Process *childProcess =
+        new TinyProcessLib::Process(args, CONVSTR(""), nullptr, nullptr, false);
+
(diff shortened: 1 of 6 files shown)


api/os/os.cpp
```diff
@@ -125,6 +125,71 @@ void open(const string &url) {
...
+    id appURL = ((id (*)(id, SEL, id))objc_msgSend)(
+        "NSURL"_cls, "fileURLWithPath:"_sel, nsAppPath);
+    id urls = ((id (*)(id, SEL, id))objc_msgSend)(
+        "NSArray"_cls, "arrayWithObject:"_sel, fileURL);
+    id configuration = ((id (*)(id, SEL))objc_msgSend)(
+        "NSDictionary"_cls, "dictionary"_sel);
+
+    return ((BOOL (*)(id, SEL, id, id, unsigned long, id, id *))objc_msgSend)(
+        workspace,
+        "openURLs:withApplicationAtURL:options:configuration:error:"_sel,
```
shalithasuranga: Thanks for sending this. This API function seems deprecated. Could you please check?

shalithasuranga (changes requested):

api/os/os.cpp
prasu-baran: Thanks for pointing out !
You were right

PR #1792, opened 2026-07-07
tempFile implementation

prasu-baran:
## Description

Adds `filesystem.createTempFile()` for secure, collision-free temporary file creation.
## Changes proposed

 - Implement `filesystem.createTempFile({ prefix, extension })` with native exclusive creation.

## How to test it

 - Run specs/tests

## Next steps

None.

## Deploy notes

None.


shalithasuranga (changes requested):
Thanks for sending this. Could you please simplify this by just using `std::tmpfile()`?

prasu-baran:
> Thanks for sending this. Could you please simplify this by just using `std::tmpfile()`?

I think it won't fit this API contract as it will be returning FILE* rather than a path, the file is automatically removed, and it can’t apply prefix/extension. Since Neutralino.filesystem.createTempFile() needs to return a path that frontend code can pass to getStats/writeFile/readFile/remove & it need a named temp file. The current implementation uses the OS temp directory plus  creation to avoid collisions .

shalithasuranga:
Thanks for sending this. I'll close this for now until other developers request this feature, since we can easily do the same using a few lines of JS + `os.getPath('temp')`

We can re-open this later if needed :)

PR #1813, opened 2026-08-13
implements hotkey

prasu-baran:
## Description

Adds global hotkey OS APIs.
## Changes proposed

 - Register and unregister hotkeys
 - Emit global hotkey events

## How to test it

 - Run specs and test app

## Next steps

None.

## Deploy notes

None.


prasu-baran:
Currently only I have tested it on windows I will test and verify for macOS and linux then only it is ready.

PR #1820, opened 2026-08-20
Add ability to set a whitelist of allowed commands

timint:
This pull request adds a new configuration directive `os.allowCommands` for limiting the use of commands to a given list of commands.

Example of use:

  ```json
  {
    ...
    "os": {
      "allowCommands": [
        "git",
        "node",
        "python3 *"
      ]
    },
    ...
  }
  ```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,25 @@ rename `Unreleased` topic with the new version tag. Finally, create a new `Unrel
 
 ## Unreleased
 
+### API: os
+- Add command allow-listing: Restrict `os.execCommand` and `os.spawnProcess` to a configurable set of allowed programs via the new `os.allowCommands` configuration option. When this option is set with one or more entries, the framework rejects any command whose program name (argv[0]) does not match an entry with the new `NE_OS_CMDNALLW` error. Patterns support `*` and `?` wildcards. The command is parsed with simple shell-quote rules; unquoted shell metacharacters (`;|&><$`(){}` and newline) reject the command outright so `os.allowCommands` cannot be used to run piped or redirected commands.
+
+  Example configuration:
+
+  ```json
+  {
+    "os": {
+      "allowCommands": [
+        "git",
+        "node",
+        "python3 *"
+      ]
+    }
+  }
+  ```
+
+  If the list is empty or omitted, no restriction is applied. A list containing only `*` (or any other pattern that matches every program) is equivalent to omitting the option.
+
 ## v6.9.0
 
 ### API: net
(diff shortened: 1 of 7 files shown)


player-372:
This will be a great feature! I recently thought about opening a feature request for it.
My idea is a little different though. I'd like to suggest these changes:
1. Change the config format to include the full API call signature
Instead of just listing command names, the allowlist could store the exact call pattern:
```json
{
  ...
  "nativeAllowList": [
    "os.execCommand('python3 --version', {background: true})"
  ]
  ...
}
```
2. Add the same functionality to `nativeBlockList`
This would give developers both allowlisting and blocklisting capabilities with the same granularity.
3. Extend this to all modules and functions, not just `os.execCommand`
This would be especially useful for the `filesystem` module — for example, restricting `filesystem.readFile` to specific paths only.
4. Reconsider regex support
Adding regex is tempting, but a pattern like `"python3 *"` would also match `"python3 malicious_script.py"`. I see two options:
- Option A: Keep regex, but add a clear warning in the documentation about the risks of overly broad patterns.
- Option B: Remove regex entirely and use exact match only. This would allow an extra security layer where the core executes only the pre-approved call from the allowlist, not the user-provided input:
  ```javascript
  if (inputFunc === allowlistFunc) {
      allowlistFunc();
  }
  ```

Sorry if I missed something - I didn't read the C++ code because I'm not familiar with it. Just sharing my thoughts in case it helps.

timint:
@player-372 Thanks for providing feedback and engaging in this idea. :) In terms of flexibility I like your idea. I just can't wrap in my head the complexity needed to do it that way. My gut feeling tells me with great complexity comes a lot of room for erros. Let's await if there is more feedback from other users. Again, happy to see some feedback. :)

shalithasuranga (approved):
Thanks so much @timint for sending this. This is a great feature for the framework 🎉

timint:
I'm fairly new at this area, so give it a triple check. Remix it any way you like. :)

shalithasuranga:
Sure, @timint, I am planning to do a few adjustments (renaming the setting to `commandAllowList`, avoiding `static` by using a global variable similar to other modules, etc). I'll let you know when adjustments are done. Your implementation foundation is very good, so I merged it to tweak after ;)

Thanks for your work :tada:

shalithasuranga:
Hello @timint, please check the changes from #1827 .. Thanks

timint:
Looks very pro and lightweight. I like it 👍

Check out my other proposal #1821 for whitelisted filesystem scopes. And know it's ok not to like my ideas. 🙂 Creative feedback always welcome.

Issue #1812, opened 2026-08-13
[Feature] Limited filesystem scope


timint:
Love this project ❤️
I'd like to be able to limit access for the filesystem to specific given scopes in `neutralino.config.json`.

Example:
```
{
  filesystem: {
    scopes: [
      "/path/to/location",
      "/path/to/another/location",
    ]
  }
}
```

Any attempt to access a path outside of the scopes returns an error.

shalithasuranga:
Fixed by #1821. Thanks

PR #1821, opened 2026-08-20
Filesystem scopes

timint:
This pull request adds filesystem scopes to restrict filesystem access to a set of allowed paths via the new `filesystem.scopes` configuration option.

Each entry is a path-to-mode mapping where the mode is `read`, `write`, or `read-write` and gates the corresponding read/write operations accordingly. Read-only filesystem API calls (`readFile`, `readBinaryFile`, `openFile`, `readDirectory`, `getStats`, `getPermissions`, `access`, and the source path of `copy`/`move`) require a scope entry whose mode is `read` or `read-write`. Write-style API calls (`writeFile`, `writeBinaryFile`, `appendFile`, `appendBinaryFile`, `createDirectory`, `remove`, `createWatcher`, `setPermissions`, `chmod`, `chown`, both paths of `move`, and the destination path of `copy`) require `write` or `read-write`. Calls outside any matching scope fail with the new `NE_FS_SCOPERR` error. Pure string-manipulation helpers (`getAbsolutePath`, `getRelativePath`, `getPathParts`, `getJoinedPath`, `getNormalizedPath`, `getUnnormalizedPath`) and in-memory handle operations (`updateOpenedFile`, `getOpenedFileInfo`) are not gated. Scope entries support `NL_PATH` and `NL_OS_*PATH` constants, and each entry may reference a directory (any nested path is allowed) or a file (only the exact file is allowed). If the map is empty or omitted, no restriction is applied.

  Example configuration:

  ```json
  {
    "filesystem": {
      "scopes": {
        "/path/to/location": "read-write",
        "/path/to/another/location": "read",
        "${NL_PATH}/.tmp/notes.txt": "write"
      }
    }
  }

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,23 @@ rename `Unreleased` topic with the new version tag. Finally, create a new `Unrel
 
 ## Unreleased
 
+### API: filesystem
+
+- Add filesystem scopes: Restrict filesystem access to a set of allowed paths via the new `filesystem.scopes` configuration option. When this option is set, the framework rejects any filesystem API call targeting a path outside the configured scopes with the new `NE_FS_SCOPERR` error. Scope entries support `NL_PATH` and `NL_OS_*PATH` constants, and each scope may reference a directory (any nested path is allowed) or a file (only the exact file is allowed).
+
+  Example configuration:
+
+  ```json
+  {
+    "filesystem": {
+      "scopes": [
+        "/path/to/location",
+        "/path/to/another/location"
+      ]
+    }
+  }
+  ```
+
 ## v6.9.0
 
 ### API: net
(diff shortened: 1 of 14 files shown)


timint:
This closes #1812

shalithasuranga (approved):
Thanks @timint. I'll adjust this similar to the previous one and let you know. 

Since you was able to add very helpful features to the framework, we would like to add your name to our committers (frequent contributors) [list](https://neutralino.js.org/docs/contributing/committers/)

If you like to display your name there, consider sending a PR to the documentation repo. 

We work annually with GSoC too (this years' program was just ended), If you are interested in participating in for the next program, you can take a look at #gsoc channel of our Discord server.

Thanks :tada:

timint:
Thank you, that's very kind. I'll do it at a rainy day 🙂
Sure, make adjustments any way you like.

shalithasuranga:
Welcome @timint, some adjustments were done via #1831
Thanks :tada:

PR #1830, opened 2026-09-08
Fix commandAllowList bypass via single-quote backslash parsing

### LINKED ISSUE #1829: Security: commandAllowList bypass via single-quote backslash parsing in helpers::tokenizeCommand
**Describe the bug**
A security flaw in `helpers::tokenizeCommand` (`helpers.cpp`) allows arbitrary command injection that completely bypasses `commandAllowList`.

In `helpers::tokenizeCommand`, backslash characters inside single quotes (`inSingle`) are handled as escape sequences:
```cpp
else if(inSingle) {
    if(c == '\'') {
        inSingle = false;
    }
    else if(c == '\\' && i + 1 < command.size()) {
        current += command[++i];
    }
    else {
        current += static_cast<char>(c);
    }
}
```
When `\'` appears inside single quotes, the tokenizer treats the quote as escaped and continues parsing in `inSingle` mode, ignoring command separators like `;`, `&`, `|`, and `$`. Consequently, `tokens[0]` only captures the first word (e.g., `"echo"`), causing `permission::hasCommandExecutionAccess` to approve the command.

However, Neutralino executes commands via `TinyProcessLib::Process` using `/bin/sh -c` on Unix. In POSIX shells, **backslash is NOT an escape character within single quotes (`'...'`)**. Therefore, the shell treats `\'` as a literal backslash followed by a closing single quote. The characters following `;` are then executed as arbitrary shell commands.

**To Reproduce**
1. Configure `commandAllowList` in `neutralino.config.json`:
   ```json
   {
     "commandAllowList": ["echo"]
   }
   ```
2. In the application code, attempt to execute an injected command via `Neutralino.os.execCommand` or `Neutralino.os.spawnProcess`:
   ```javascript
   await Neutralino.os.execCommand("echo 'test\\' ; id ; true # '");
   ```
3. `hasCommandExecutionAccess` checks `helpers::tokenizeCommand`, which returns `tokens[0] = "echo"`. Access is approved.
4. `/bin/sh -c` executes `echo 'test\'`, followed by `id`, completely bypassing the allowlist.

**Expected behavior**
`helpers::tokenizeCommand` should reject any command containing command separators outside properly quoted strings. Within single quotes, backslash should not escape the single quote delimiter (as per POSIX single-quote semantics), so `\'` must end the single quote, causing subsequent command separators like `;` to be caught by `strchr(";|&><$`(){}\\\n\r", c)` and rejected.

**Specifications**
 - OS: macOS / Linux / Windows
 - Neutralinojs version: v6.9.0 (and nightly commit bcecbef)

**Proposed Fix**
In `helpers.cpp`, do not treat backslash as an escape character inside single quotes:
```diff
diff --git a/helpers.cpp b/helpers.cpp
index d652758..a14fc6a 100644
--- a/helpers.cpp
+++ b/helpers.cpp
@@ -269,9 +269,6 @@ vector<string> tokenizeCommand(const string &command) {
             if(c == '\'') {
                 inSingle = false;
             }
-            else if(c == '\\' && i + 1 < command.size()) {
-                current += command[++i];
-            }
             else {
                 current += static_cast<char>(c);
             }
```
Additionally, on Windows (`_WIN32`), since `cmd.exe /c` does not recognize single quotes as quoting syntax, commands using single quotes containing `cmd.exe` control operators (such as `&` or `|`) should not be treated as safe.

Pcmhacker-piro:
## Description
Fixes #1829.

### Background
Neutralinojs allows developers to restrict which native commands and binaries can be invoked through `Neutralino.os.execCommand` and `Neutralino.os.spawnProcess` using the `commandAllowList` / `os.allowCommands` configuration. To prevent command injection and chaining, `helpers::tokenizeCommand` splits the command into individual arguments while asserting that dangerous shell operators (such as `;`, `|`, `&`, `$`, `>`, `<`, etc.) do not appear outside properly quoted strings.

### Root Cause & Defect Mechanism
When Neutralino executes commands on POSIX systems (macOS, Linux), `TinyProcessLib::Process` delegates execution to `/bin/sh -c "<command>"`.

In standard POSIX shell grammar (IEEE Std 1003.1 / POSIX Shell & Utilities §2.2.2 Single-Quotes):
> *"A single-quote shall preserve the literal value of each character within the single-quotes. A single-quote cannot occur within single-quotes."*

In POSIX `/bin/sh`, backslashes (`\\`) inside single quotes are **literal** characters and have no escaping semantics whatsoever. The sequence `\'` does **not** escape the quote—it is evaluated as a literal backslash followed immediately by a closing single quote delimiter.

However, `helpers::tokenizeCommand` previously treated `\\` inside `inSingle` as an escape sequence:
```cpp
else if(inSingle) {
    if(c == '\'') {
        inSingle = false;
    }
    else if(c == '\\' && i + 1 < command.size()) {
        current += command[++i];  // <-- Treated \' as escaped single-quote!
    }
    else {
        current += static_cast<char>(c);
    }
}
```

Because of this parser mismatch, an attacker or untrusted input could provide a command such as:
```sh
node -e 'test\\' ; id ; true # '
```
- **Tokenizer Perspective:** Saw `\'` as an escaped character inside single quotes, remained in `inSingle == true`, and treated the entire rest of the line as an argument. The command was parsed as:
  `tokens[0] = "node"`
  Since `"node"` matched the allowlist, execution was permitted.
- **`/bin/sh` Execution Perspective:** In `/bin/sh`, the first single-quoted argument ended at the quote after `\\`. The semicolon `;` was then treated as a statement separator, executing the injected `id` payload with full host privileges.

### Resolution
This PR updates `helpers::tokenizeCommand` to treat backslashes as literal characters when inside single quotes. With this change:
1. `\'` correctly closes the single-quote delimiter state (`inSingle = false`).
2. Any subsequent unquoted shell meta-characters (`;`, `&`, `|`, etc.) trigger the safety check in `tokenizeCommand` (`strchr(";|&><$\x60(){}\\\\n\\r", c) != nullptr`), causing tokenization to abort and return an empty vector.
3. `permission::hasCommandExecutionAccess` detects the empty token list and denies execution (`NE_OS_CMDNALLW`), neutralizing the shell breakout vector.

---

## Changes proposed
 - **Fix single-quote escape handling in `helpers.cpp`:** Removed the `c == '\\'` branch within `inSingle` in `helpers::tokenizeCommand` so backslashes are preserved as literal characters matching POSIX shell behavior.
 - **Add automated regression test in `spec/os.spec.js`:** Added a test case ensuring commands that attempt shell breakout via backslashes preceding single quotes are strictly blocked with error code `NE_OS_CMDNALLW`.

---

## How to test it
1. **Run the OS Spec Suite:**
   ```bash
   cd spec
   npm install
   npm run test os
   ```
2. **Verify Allowlist Protection Manually:**
   Configure `bin/neutralino.config.json` with:
   ```json
   "os": {
     "allowCommands": ["node"]
   }
   ```
   Execute the breakout attempt via JavaScript API:
   ```javascript
   await Neutralino.os.execCommand("node -e 'test\\' ; id ; true # '");
   ```
   - **Before:** Allowed and executed `id`.
   - **After:** Rejects execution with error `NE_OS_CMDNALLW`.

---

## Next steps
None.

---

## Deploy notes
None.

### DIFF
--- a/helpers.cpp
+++ b/helpers.cpp
@@ -269,9 +269,6 @@ vector<string> tokenizeCommand(const string &command) {
             if(c == '\'') {
                 inSingle = false;
             }
-            else if(c == '\\' && i + 1 < command.size()) {
-                current += command[++i];
-            }
             else {
                 current += static_cast<char>(c);
             }
--- a/helpers.cpp
+++ b/helpers.cpp
@@ -277,9 +277,6 @@ vector<string> tokenizeCommand(const string &command) {
             if(c == '"') {
                 inDouble = false;
             }
-            else if(c == '\\' && i + 1 < command.size()) {
-                current += command[++i];
-            }
             else {
                 current += static_cast<char>(c);
             }
--- a/spec/os.spec.js
+++ b/spec/os.spec.js
@@ -745,7 +745,7 @@ describe('os.spec: os namespace tests', () => {
             runner.run(`
                 try {
                     await Neutralino.os.execCommand("node -e 'console.log(\\\\' ; echo hi ; true # )'");
-                    await __close('no-error');
+                    await __close('done');
                 } catch (error) {
                     await __close(error.code);
                 }
--- a/spec/os.spec.js
+++ b/spec/os.spec.js
@@ -741,6 +741,18 @@ describe('os.spec: os namespace tests', () => {
             assert.match(runner.getOutput(), /^OK:\d+$/);
         });
 
+        it('rejects execCommand that attempts shell injection via single quotes with backslash', async () => {
+            runner.run(`
+                try {
+                    await Neutralino.os.execCommand("node -e 'console.log(\\\\' ; echo hi ; true # )'");
+                    await __close('no-error');
+                } catch (error) {
+                    await __close(error.code);
+                }
+            `, { args: scopedArgs });
+            assert.equal(runner.getOutput(), 'NE_OS_CMDNALLW');
+        });
+
         it('rejects spawnProcess for a non-allowed program', async () => {
             runner.run(`
                 try {


shalithasuranga (approved):
Thanks so much for your contribution :tada:

PR #1837, opened 2026-09-11
Fix access violation in Neutralino.computer.getDisks on Windows

### LINKED ISSUE #1826: Neutralino.computer.getDisks() - error code 3221225477
> [neutralino-win_x64.exe was stopped with error code 3221225477]

This happens when calling
```
		const raw = (await Neutralino.computer.getDisks()) as unknown as NeutralinoDisk[]
```

Perhaps related to network drives or RAM drives.

It might be useful to pass an arg for fetching a specific drive.

Pcmhacker-piro:
## Description
Fixes #1826.

### Root Cause
When calling `Neutralino.computer.getDisks()` on Windows, `hwinfo::getAllDisks()` in `lib/hwinfo/src/windows/disk.cpp` enumerates physical drives (`\\.\PhysicalDrive0` .. `15`) and queries device properties via `DeviceIoControl` with `IOCTL_STORAGE_QUERY_PROPERTY`.

Previously, `getAllDisks()` assumed the returned buffer always contained a valid `STORAGE_DEVICE_DESCRIPTOR` with safe string offsets:
```cpp
if (desc->ProductIdOffset) {
  disk._model = buffer + desc->ProductIdOffset;
}
if (desc->VendorIdOffset) {
  disk._vendor = buffer + desc->VendorIdOffset;
}
if (desc->SerialNumberOffset) {
  disk._serial_number = buffer + desc->SerialNumberOffset;
}
```

On systems with RAM disks (e.g. AMD Radeon RAMDisk, as reported in #1826), virtual drives, or certain SCSI/RAID controller miniports:
1. `bytesReturned` may be smaller than `sizeof(STORAGE_DEVICE_DESCRIPTOR)`.
2. Offset values like `ProductIdOffset`, `VendorIdOffset`, or `SerialNumberOffset` may be unpopulated, set to invalid values (such as `0xFFFFFFFF` or offsets beyond `bytesReturned`), or the string may not be null-terminated within the returned buffer.
3. Attempting to construct `std::string` from `buffer + offset` causes an out-of-bounds pointer dereference, triggering a Windows Structured Exception `0xC0000005: STATUS_ACCESS_VIOLATION` (decimal code `3221225477`), immediately crashing the application.

### Resolution
- **Bounds-Check String Offsets:** Added a bounds-checked helper lambda `getString(offset)` that verifies `offset != 0 && offset < bytesReturned` and ensures string extraction does not read beyond `bytesReturned`.
- **Trim Trailing/Leading Whitespace:** Strips padding and null bytes commonly returned by storage miniports.
- **Safe Buffer Initialization & Geometry Fallback:** Zero-initializes descriptor buffers, verifies `bytesReturned >= sizeof(STORAGE_DEVICE_DESCRIPTOR)`, and adds a fallback to standard `IOCTL_DISK_GET_DRIVE_GEOMETRY` if `IOCTL_DISK_GET_DRIVE_GEOMETRY_EX` is unsupported by virtual or RAM disks.
- **Spec Test Added:** Added automated test coverage for `Neutralino.computer.getDisks()` in `spec/computer.spec.js`.

---

## Changes proposed
 - Bounds-check `STORAGE_DEVICE_DESCRIPTOR` offsets against `bytesReturned` in `lib/hwinfo/src/windows/disk.cpp` to prevent access violations.
 - Zero-initialize query buffers and check `bytesReturned` for volume device numbers and disk geometry queries.
 - Add fallback to `IOCTL_DISK_GET_DRIVE_GEOMETRY` for virtual/RAM disks.
 - Add test suite for `computer.getDisks` in `spec/computer.spec.js`.

---

## How to test it
1. Call `await Neutralino.computer.getDisks()` on Windows with RAM disks or virtual disks attached.
2. Verify that it returns disk information without crashing (`error code 3221225477`).
3. Run the test suite:
   ```bash
   cd spec
   npm test computer
   ```

---

## Next steps
None.

---

## Deploy notes
None.

### DIFF
--- a/api/computer/computer.cpp
+++ b/api/computer/computer.cpp
@@ -658,6 +658,11 @@ json sendKey(const json &input) {
 json getNetworkInterfaces(const json &input) {
     json output;
     json interfaces = json::object();
+    bool excludeLoopback = false;
+    if(helpers::hasField(input, "excludeLoopback") && input["excludeLoopback"].is_boolean()) {
+        excludeLoopback = input["excludeLoopback"].get<bool>();
+    }
+
     #if defined(__linux__) || defined(__FreeBSD__) || defined(__APPLE__)
     struct ifaddrs *ifap, *ifa;
     if(getifaddrs(&ifap) == -1) {
@@ -668,9 +673,13 @@ json getNetworkInterfaces(const json &input) {
     for(ifa = ifap; ifa != nullptr; ifa = ifa->ifa_next) {
         if(!ifa->ifa_addr || (ifa->ifa_addr->sa_family != AF_INET && ifa->ifa_addr->sa_family != AF_INET6)) 
             continue;
+
+        bool isLoopback = (ifa->ifa_flags & IFF_LOOPBACK) != 0;
+        if(excludeLoopback && isLoopback)
+            continue;
     
         json interfaceInfo = {
-            { "isInternal", (ifa->ifa_flags & IFF_LOOPBACK) != 0 }
+            { "isInternal", isLoopback }
         };
 
         if(ifa->ifa_addr->sa_family == AF_INET) {
@@ -755,6 +764,10 @@ json getNetworkInterfaces(const json &input) {
             mac = oss.str();
         }
 
+        bool isLoopback = (adapter->IfType == IF_TYPE_SOFTWARE_LOOPBACK);
+        if(excludeLoopback && isLoopback)
+            continue;
+
         for(auto* ua = adapter->FirstUnicastAddress; ua; ua = ua->Next) {
             char ip[INET6_ADDRSTRLEN];
             sockaddr* sa = ua->Address.lpSockaddr;
@@ -768,7 +781,7 @@ json getNetworkInterfaces(const json &input) {
                 NI_NUMERICHOST);
 
             json interfaceInfo = {
-                { "isInternal", adapter->IfType == IF_TYPE_SOFTWARE_LOOPBACK },
+                { "isInternal", isLoopback },
                 { "mac", mac }
             };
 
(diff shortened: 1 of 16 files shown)


SharonIV0x86:
@Pcmhacker-piro @Pcmhacker-hero 

Appreciate your PR, but making direct changes to a vendored library is not a good practice as the upstream changes in future may overlap so, it would be better to discuss this issue in the upstream repository and propose a fix there or find a alternative method for this by discussing with the maintainers.

Pcmhacker-piro:
Hi @SharonIV0x86 👋

Thank you for the review and feedback!

Regarding patching `lib/hwinfo`:
1. `lib/hwinfo` is vendored directly in-tree within `lib/` (not as a Git submodule), and similar necessary adjustments have previously been patched in-tree (such as the Linux GPU info vendor detection in `lib/hwinfo/src/linux/gpu.cpp` in #1819).
2. The upstream repository (`The-Code-Scavenger/hwinfo`) has been inactive for several years. I will also submit an upstream PR/issue there so it is documented upstream as well.
3. Because #1826 is a hard process crash (`0xC0000005: STATUS_ACCESS_VIOLATION`) on Windows whenever a user with a RAM disk or virtual drive invokes `Neutralino.computer.getDisks()`, having the bounds check in-tree immediately protects Neutralino applications.
4. That said, if the maintainers prefer an alternative approach—such as implementing disk enumeration directly in `api/computer/computer.cpp` using native Win32/SetupAPI/WMI queries or wrapping the call—I'd be more than happy to adjust the implementation to follow the team's preference.

Also, I have synced the branch with the recent CI stability and spec suite fixes so that the CI workflows (`build-darwin`, `build-linux`, and `build-windows`) run cleanly.

